PR #10, opened 2021-06-10
Fix one typo found during documentation review

juadk:
harvester/harvester#833
I've found this only one typo during the documentation review.


guangbochen (approved):
LGTM

PR #14, opened 2021-06-23
Add write_files in the Harvester configuration

juadk:
Add the new `os.write_files` feature to the documentation.


guangbochen (approved):
LGTM

PR #26, opened 2021-09-29
Fix bond_option field name in networks

bk201:
Related issue: https://github.com/harvester/harvester/issues/1323


weihanglo (approved):
My bad. Thanks for the fix!

guangbochen (approved):
LGTM

PR #47, opened 2021-10-13
Add bond_options and VLAN network example to the PXE installation section

bk201:
**Related issue**
https://github.com/harvester/harvester/issues/1347

### DIFF
--- a/docs/install/pxe-boot-install.md
+++ b/docs/install/pxe-boot-install.md
@@ -76,11 +76,21 @@ os:
 install:
   mode: create
   networks:
-    harvester-mgmt:       # The management bond name. This is mandatory.
+    harvester-mgmt:       # (Mandatory) The management bond name.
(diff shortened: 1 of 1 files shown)


docs/install/pxe-boot-install.md
```diff
@@ -122,11 +133,21 @@ os:
...
   networks:
-    harvester-mgmt:       # The management bond name. This is mandatory.
+    harvester-mgmt:       # (Mandatory) The management bond name.
       interfaces:
       - name: ens5
       default_route: true
       method: dhcp
+      bond_options:
+        mode: balance-tlb
+        miimon: 10
```
weihanglo, line 143: These are the defaults. Should we list them here or just leave them empty?

weihanglo: Fair. Thanks.

PR #61, opened 2021-10-15
Add a workaround to disable overlaid directories in debugrw mode

bk201:
This is needed before https://github.com/harvester/os2/pull/18 is in.


guangbochen (approved):
LGTM

PR #82, opened 2021-12-01
Add config: 'no_data_partition'

johnliu55tw:
- Add doc for `no_data_partition` config.

- Remove `force_mbr` from `os` section.

- Reference `no_data_partition` from `force_mbr` config.

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -48,7 +48,6 @@ os:
   labels:
     foo: bar
     mylabel: myvalue
-  force_mbr: false
 install:
   mode: create
   networks:
@@ -67,6 +66,8 @@ install:
   vip: 10.10.0.19
   vip_hw_addr: 52:54:00:ec:0e:0b
   vip_mode: dhcp
+  force_mbr: false
+  no_data_partition: false
 system_settings:
   auto-disk-provision-paths: ""
 ```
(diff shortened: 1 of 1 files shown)


docs/install/harvester-configuration.md
```diff
@@ -441,3 +442,18 @@ install:
   vip_mode: dhcp
   vip_hw_addr: 52:54:00:ec:0e:0b
 ```
+
+### `install.no_data_partition`
+
+#### Definition
+
+Do not create an additional disk partition for storing VM data
+and use operating system partition to store them. Default: `false`.
```
nunix, line 451:
```suggestion
Do not create an additional disk partition for storing VM data. The operating system partition will be used to store the VM data. 

Default: `false`.
```

guangbochen, line 451: Do not create additional disk partitions for VM data, the system partition will be used by default.
Can't be resized or reverted after the installation, useful when users want to use additional disks for the VM data with [auto-provision-path]() config. Default: `false`.

nunix (commented):
rephrased the description

docs/install/harvester-configuration.md
johnliu55tw: Rephrased it a bit based on your suggestion.

johnliu55tw: Rephrased it a bit based on your suggestion.

johnliu55tw:
@guangbochen @nunix I think we need to merge this PR first because I've fixed some minor issues, then I will rebase my other PRs based on this.

nunix (approved):

guangbochen (approved):

PR #94, opened 2021-12-20
Add vm-force-reset-policy setting

FrankYang0529:


docs/settings/settings.md
```diff
@@ -282,3 +282,18 @@ The following example sets the enabled SSL/TLS protocols to `TLSv1.2` and `TLSv1
   "ciphers": "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305"
 }
 ```
+
+## `vm-force-reset-policy`
+
+This setting allows you to force to reschedule VM when a node is unavailable. When a node turns from Ready to Not Ready, the controller will wait `period` seconds. If a node is still Not Ready, it will force delete the pod of the VM on that node and reschedule it to another available node.
```
guangbochen, line 288: This setting allows you to force reschedule VMs when a node is unavailable. When a node turns to be `Not Ready`, it will force delete the VM on that node and reschedule it to another available node after a period of seconds.

FrankYang0529: Updated it. Thank you.

guangbochen (approved):
LGTM

PR #110, opened 2022-01-12
Remove extra tab character

nunix:
Removed an extra `tab` that breaks the page rendering


nunix:
@guangbochen this only applies to the `create-windows-vm.md` page

guangbochen (approved):
LGTM, can u please help to resolve the conflict, thanks.

PR #118, opened 2022-02-10
Fixed Code block malformed from last commit

belgaied2:
Due to the last commit in the docs, the code block closing is broken. GitHub shows the code block correctly but not the generated website.

This pull request fixes that by adding the necessary code blocking closing.


guangbochen (approved):
LGTM

PR #121, opened 2022-03-03
Remove the statement using DHCPd-provided hostname if a hostname is not set

bk201:
**Related issue**

https://github.com/harvester/harvester/issues/1444#issuecomment-1057604626


guangbochen (approved):
LGTM

PR #126, opened 2022-03-14
Added ISO images info to Harvester node driver section

jtravee:
Added blurb on new ISO images in Rancher UI; see [rancher/docs issue for reference](https://github.com/rancher/docs/pull/3948).


jtravee:
@guangbochen, I placed this note in the `node driver` page to be more noticeable for users. But please let me know if there is another section/page in which you would prefer that I would make this note.

guangbochen (approved):

PR #141, opened 2022-04-21
fix 1842, add memory usage formula explanation

w13915984028:
related issues:
https://github.com/harvester/harvester/issues/1842

Add explanation of the memory usage formula into Harvester docs/monitoring/monitoring.md.


guangbochen (approved):
LGTM

PR #143, opened 2022-05-02
fix-2211 Add description of monitoring resource setting in earlier versions and node-exporter setting

w13915984028:
Add description of monitoring resource setting in earlier versions, the guide of CLI will help user to set proper values in Harvester v1.0.0 or v0.3.0.

The value setting of node-exporter is available in v1.0.2, is also added to documents.

related issue:
https://github.com/harvester/harvester/issues/2211
https://github.com/harvester/harvester/issues/2075


w13915984028:
The feature in
https://github.com/harvester/harvester/issues/2075
is also updated into docs in this PR.

guangbochen (approved):
LGTM

Issue #144, opened 2022-05-03
Docs suggest more than 3 nodes are required for HA


himslm01:
https://github.com/harvester/docs/blob/main/docs/host/host.md says:

> When there are more than three nodes, the two other nodes that first joined are automatically promoted to management nodes to form a HA cluster.

The important point is "more than three nodes". Therefore having exactly three nodes will not be HA.

Alejandro Bonilla says that is wrong:

> If we have 3 nodes, we provide an HA stack.
> Any scenario where you have 3 nodes and they are not all cp/etcd/worker it's a bug.

https://rancher-users.slack.com/archives/C01GKHKAG0K/p1651598291509279?thread_ts=1650197492.586649&cid=C01GKHKAG0K

For that scenario, the documentation should say "When there are three or more nodes,..."

This issue is related to:
https://github.com/harvester/harvester/issues/2194

bk201:
@himslm01, thanks. This should be addressed in https://github.com/harvester/docs/pull/145

divya-mohan0209:
@himslm01 , closing this off since the issue seems to have been addressed with #145 by @bk201 . Please re-open the issue if the phrasing still seems to be an issue.

PR #145, opened 2022-05-04
Fix wrong statement about node promotion

bk201:
The original statement suggests a promotion only happens when nodes count >= 4.

https://github.com/harvester/docs/issues/144


weihanglo (approved):
LGTM

guangbochen (approved):

PR #154, opened 2022-05-13
Fix upgrade error and vm broken links

guangbochen:
Fix upgrade error and vm broken links


bk201 (approved):
Thanks. Just check again the [published artifact ](https://releases.rancher.com/harvester/v1.0.2/version.yaml)is correct. Just need to revise the doc.

PR #158, opened 2022-05-24
Fixed broken documentation links

simonflood:


simonflood:
Will redo PR with relative links

PR #171, opened 2022-05-30
Need sudo for timedatectl status command

simonflood:
otherwise get "Failed to query server: Access denied" error


guangbochen (approved):
LGTM, thanks.

PR #180, opened 2022-07-21
Add installation requirements

yaocw2020:
Related issue: https://github.com/harvester/harvester/issues/2216


bk201 (approved):
LGTM
Rancher pods also open host ports `8443`, should we add it too?

divya-mohan0209 (approved):
lgtm!

guangbochen (approved):

yaocw2020:
> LGTM Rancher pods also open host ports `8443`, should we add it too?

It seems the port is inside the pod, not open in the host.

PR #182, opened 2022-07-25
Updating the container image

clemenko:
Updating the container image to registry.suse.com/suse/vmdp/vmdp:2.5.3 as per the default template.


guangbochen (approved):
LGTM and thanks for the PR.

PR #184, opened 2022-08-04
Network Hardware Interface MAC Support

irishgordo:
**Reason:**
Show that through the configuration on a Harvester Configuration file that you can specify the interface's MAC address.
Since within the source code, we do allow for it via: 
https://github.com/harvester/harvester-installer/blob/v1.0.3-rc2/pkg/config/config.go#L18

**Tested Locally:**
show-in-docs-you-can-add-hwaddr

Resolves: docs/harvester-config-with-hwaddr-for-network-interfaces

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -54,6 +54,7 @@ install:
     harvester-mgmt:
       interfaces:
       - name: ens5
+        hwAddr: "B8:CA:3A:6A:64:7C"
       method: dhcp
   force_efi: true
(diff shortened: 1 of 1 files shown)


noahgildersleeve (approved):
LGTM

noahgildersleeve:
@irishgordo You might mark it as optional in the code with a comment the way that some of the other fields are marked as mandatory.

irishgordo:
Ahh, yeah I'll make that edit :smile:  - that's a good idea

PR #187, opened 2022-08-05
Introduce Info On Harvester Configuration Priority

irishgordo:
* a small note blurb added about configuration priority

Resolves: docs/specify-configuration-priority

**Tested Locally:**
update-configuration-priority-info

We may want to change the wording a bit, but I just wanted to scaffold this PR as an idea of what we might say that could convey that information :smile:  - super willing to change the text to something that might work better :+1: :smile: !


noahgildersleeve (approved):
LGTM

bk201:
Thanks @irishgordo !

PR #206, opened 2022-10-20
Update v1.1 path and localize search-related strings

vickyhella:
- Update the v1.1 doc path from `next` to `v1.1`
- Add translation for Algolia search-related strings

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -40,6 +40,7 @@ const config = {
           versions: {
             current: {
               label: 'v1.1-dev 🚧',
+              path: "v1.1",
             },
             "v1.0": {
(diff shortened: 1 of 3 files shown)


vickyhella:
> 

These translations are actually auto-generated by running the `yarn run write-translations -- --locale zh` command introduced in the [Docusaurus documentation](https://docusaurus.io/docs/i18n/tutorial#translate-plugin-data). I am not sure why our site already have the translation, but I think it's better that we follow the document and add those translations to the file.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
Thanks!

PR #207, opened 2022-10-20
Update Chinese translation

vickyhella:
Update Chinese translation for the `install/harvester-configuration` page


guangbochen (approved):
LGTM, thanks.

PR #222, opened 2022-10-25
Set v1.1 to the default version

guangbochen:
**Description:**
Set v1.1 to the default version


vickyhella:
Also need to change v1.0 to v1.1 in the `src/pages/index.js` file so that the v1.1 docs can be the homepage

vickyhella (approved):
Please also change `v1.1-dev` to `v1.1` in `i18n/zh/docusaurus-plugin-content-docs/current.json` (in line 3). Sorry I missed this one. Otherwise, LGTM!

PR #228, opened 2022-10-26
Possibly Bump Minimum Resource CPU Spec for v1.1.0?

irishgordo:
### Description:
Based on:
https://github.com/harvester/harvester/issues/3041#issuecomment-1291748147

Should we be bumping the minimum core-count spec on CPU for testing for v1.1.0?

Super open to close this PR if this is unneeded.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,7 +50,7 @@ To get the Harvester server up and running, the following minimum hardware is re
 
 | Type | Requirements                                                                                                                                                                                               |
 |:---|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


guangbochen:
We use default 8c for the [vagrant testing](https://github.com/harvester/ipxe-examples/pull/55) and it still works, so prefer not to set a high-barrier of hardware requirements for testing purposes.

irishgordo:
sounds good :+1:  - :smile:  - I'll go ahead and close this out

bk201:
Thanks Mike.
We can keep an eye on this. I wonder if an 8-core/single-node cluster might not have enough CPU during an upgrade.

PR #229, opened 2022-10-26
Possibly Mention in a note VM-Import for OpenStack Tested On Stable/Train

irishgordo:
Possibly leave a small note to mention that OpenStack vm-import has been tested on version stable/train.
Knowing that testing will expand to other environments later.

Super open to close this PR if this is unnecessary to mention :smile:


irishgordo:
@ibrokethecloud @bk201 -
Closing this PR, as I spent some time building out an OpenStack DevStack instance at version "Zed" which is the latest "stable" release of OpenStack - and validated that we can import VMs successfully from OpenStack version Zed.

PR #232, opened 2022-10-27
fix Rancher support matrix

guangbochen:
**Description:**
Fix the Rancher v2.6.9 support matrix to include v1.0.3, in Rancher v2.6.9 we have introduced the UI plugin mechanism that will allow each Harvester cluster to serve its UI assets via the Harvester API, for older versions like Harvester v1.0.3 it will just load the embedded UI when the plugin is not available.


bk201 (approved):
LGTM!

PR #235, opened 2022-11-02
Link to SUSE instructions for the VMDP drivers

Martin-Weiss:
This also includes the qemu guest agent required to expose / show the right IP address of the guest in the Harvester UI.


guangbochen (approved):
LGTM, thanks.

PR #248, opened 2022-11-17
storagenetwork: whereabouts cni is not always installed properly.

tjjh89017:
whereabouts cni & crds are not always installed properly after harvester is upgraded from v1.0.3 to v1.1.0 or above

Related issues: https://github.com/harvester/harvester/issues/3168

### DIFF
--- a/docs/advanced/storagenetwork.md
+++ b/docs/advanced/storagenetwork.md
@@ -30,11 +30,11 @@ There are some prerequisites before configuring the Harvester Storage Network se
 
 :::caution
 
-If the Harvester cluster is upgraded from v1.0.3, please check if Whereabouts CNI is installed properly before you move on to the next step. We will always recommend following this guide to check. [Issue 3168](https://github.com/harvester/harvester/issues/3168) describes that the Harvester cluster will not always install Whereabouts CNI properly.
+If the Harvester cluster was upgraded from v1.0.3, please check if Whereabouts CNI is installed properly before you move on to the next step. We will always recommend following this guide to check. [Issue 3168](https://github.com/harvester/harvester/issues/3168) describes that the Harvester cluster will not always install Whereabouts CNI properly.
 
-- We should get `ippools.whereabouts.cni.cncf.io` without errors in the following command.
+- Verify the `ippools.whereabouts.cni.cncf.io` CRD exists with the following command.
     - `kubectl get crd ippools.whereabouts.cni.cncf.io`
-- If the Harvester cluster didn't have `ippools.whereabouts.cni.cncf.io`, please add [these two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds) before configuring `storage-network` setting.
+- If the Harvester cluster doesn't have `ippools.whereabouts.cni.cncf.io`, please add [these two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds) before configuring `storage-network` setting.
 ```
 kubectl apply -f https://raw.githubusercontent.com/harvester/harvester/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds/whereabouts.cni.cncf.io_ippools.yaml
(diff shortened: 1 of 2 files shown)


docs/advanced/storagenetwork.md
```diff
@@ -28,6 +28,20 @@ There are some prerequisites before configuring the Harvester Storage Network se
 - All pods that are attached to Longhorn Volumes should be stopped.
     - Users could skip this step with the Harvester Storage Network setting. Harvester will stop Longhorn-related pods automatically.
 
+:::caution
+
+If the Harvester cluster is upgraded from v1.0.3, please check if Whereabouts CNI is installed properly before you move on to the next step. We will always recommend following this guide to check. [Issue 3168](https://github.com/harvester/harvester/issues/3168) describes that the Harvester cluster will not always install Whereabouts CNI properly.
+
+- We should get `ippools.whereabouts.cni.cncf.io` without errors in the following command.
```
btat, line 35:
```suggestion
- Verify the `ippools.whereabouts.cni.cncf.io` CRD exists with the following command.
```

Based on the next point, it seems like the only condition to proceed is if the CRD is on the cluster. 

- If that's the case, I'd suggest being more specific so users know exactly what they're checking for.
- If not, then the introductory clause in the next point should be more generic.

docs/advanced/storagenetwork.md
```diff
@@ -28,6 +28,20 @@ There are some prerequisites before configuring the Harvester Storage Network se
 - All pods that are attached to Longhorn Volumes should be stopped.
     - Users could skip this step with the Harvester Storage Network setting. Harvester will stop Longhorn-related pods automatically.
 
+:::caution
+
+If the Harvester cluster is upgraded from v1.0.3, please check if Whereabouts CNI is installed properly before you move on to the next step. We will always recommend following this guide to check. [Issue 3168](https://github.com/harvester/harvester/issues/3168) describes that the Harvester cluster will not always install Whereabouts CNI properly.
```
btat, line 33:
```suggestion
If the Harvester cluster was upgraded from v1.0.3, please check if Whereabouts CNI is installed properly before you move on to the next step. We will always recommend following this guide to check. [Issue 3168](https://github.com/harvester/harvester/issues/3168) describes that the Harvester cluster will not always install Whereabouts CNI properly.
```

docs/advanced/storagenetwork.md
```diff
@@ -28,6 +28,20 @@ There are some prerequisites before configuring the Harvester Storage Network se
...
+
+- We should get `ippools.whereabouts.cni.cncf.io` without errors in the following command.
+    - `kubectl get crd ippools.whereabouts.cni.cncf.io`
+- If the Harvester cluster didn't have `ippools.whereabouts.cni.cncf.io`, please add [these two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds) before configuring `storage-network` setting.
```
btat, line 37:
```suggestion
- If the Harvester cluster doesn't have `ippools.whereabouts.cni.cncf.io`, please add [these two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds) before configuring `storage-network` setting.
```

Updated tense since it's the current state of the cluster being checked.

btat (changes requested):
Add a few minor suggestions.

On a sidenote, looking at the linked CRDs it looks like Whereabouts was introduced in 1.1.0. Are there cases where users on 1.0.3 would have it? If not, it may be better to just add it to the existing list of prerequisites as a new requirement rather than framing it like a bug.

bk201:
> Add a few minor suggestions.
> 
> On a sidenote, looking at the linked CRDs it looks like Whereabouts was introduced in 1.1.0. Are there cases where users on 1.0.3 would have it installed? If not and all users upgrading from 1.0.3 would need to add it then it may be better to just add it to the existing list of prerequisites as a new requirement rather than framing it like a bug.

@btat The CRDs are supposed to be there after upgrading from v1.0.3, so we frame it as a bug. We planned to fix this in v1.1.1. I added [a note](https://github.com/harvester/docs/pull/248/commits/bb956534ce6da0683ecbee92730b843d77bfb9be) in the upgrade page (v1.0.3 -> v1.1.0), please see if this looks better. Thanks!

guangbochen (approved):
LGTM, thanks.

btat (approved):

bk201 (approved):

PR #237, opened 2022-11-04
Add v1.1 translation

vickyhella:
- Add v1.1 translation

PR #253, opened 2022-11-22
Move titles to the title header and update translation

vickyhella:
- Moves the `# title` to the `title` YAML header for all docs (v0.3/v1.0/v1.1 and the corresponding Chinese docs)
- Updates Chinese translation for v1.1 
- Brings the v1.1 translation optimizations suggested by the engineering team to v1.0 translation as well
- Updates the English docs to keep the CN and EN docs consistent as mentioned in #237
- Changes `Harvester` in the `code.json` doc back to `{siteTitle}` (`{siteTitle}` is equivalent to `Harvester` in the current setting)


guangbochen (approved):
LGTM, thanks.

PR #259, opened 2022-12-02
Correct typo "additioanl-ca" to "additional-ca"

Martin-Weiss:
Correct typo "additioanl-ca" to "additional-ca"


starbops (changes requested):
I found that there's another typo in L92 in the same file, would you please update it? Thanks.

```
- Edit the `additioanl-ca` setting again. Input the certificates again and click `Save`.
```

PR #265, opened 2022-12-14
Remove VM shutting down requirement

bk201:
Shutting down is only required when upgrading from v1.0.3.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>


vickyhella (approved):
LGTM!

guangbochen (approved):
LGTM, thanks.

PR #266, opened 2022-12-15
Update Chinese translation

vickyhella:
- Updates Chinese translation, including modifying the CN translation to make it consistent with the EN docs
- Aligns formats (some, such as the indents, are done by the localization tool automatically and won't affect how it looks in production)
- Converts English punctuation to Chinese ones
- Brings content that only exists in the CN docs to the EN docs, such as the `using live migration` statement


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM! Thanks!

PR #270, opened 2022-12-23
Wrap RW mode image in OS troubleshooting

bk201:
A minor change to fix the weird layout:
From
Screenshot 2022-12-23 at 2 44 08 PM

To
Screenshot 2022-12-23 at 2 46 22 PM


vickyhella (approved):
LGTM!

PR #273, opened 2023-01-11
Fix broken links and add redirection rules

vickyhella:
- Fix broken links. Some of them are actually fine but the checker reports them. So I also modified them altoghther to avoid misleading errors.
- Add a redirection rule to redirect `/latest/install/pxe-boot-install` to `/v1.1/install/pxe-boot-install`, which is used in https://github.com/harvester/harvester README. Once this PR is merged, https://docs.harvesterhci.io/latest/install/pxe-boot-install will be redirected to https://docs.harvesterhci.io/v1.1/install/pxe-boot-install.


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

btat (approved):
LGTM

PR #283, opened 2023-02-14
Update rancher & harvester support matrix

vickyhella:
- Redirects the support matrix to the one on suse.com


guangbochen (approved):
LGTM and thanks for the fixing.

PR #293, opened 2023-03-15
#3496 remove old informations about adding network interface from install.management interface (<v1.1)

LucasSaintarbor:
Resolves [Issue #3496](https://github.com/harvester/harvester/issues/3496)

Small change - Remove the following statement, it is no longer true as of v.1.1. Users can configure network interfaces for the host machine via install.management_interface. 
- "Each key-value pair represents a network interface. The key name becomes the network name, and the values are configurations for each network."
- This is done with install.networks in <v.1.1.

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -360,9 +360,7 @@ install:
 
 #### Definition
 
-Configure network interfaces for the host machine. Each key-value pair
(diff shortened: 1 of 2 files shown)


bk201 (approved):
LGTM, thanks for the fix!

guangbochen (approved):
LGTM, thanks.

PR #288, opened 2023-03-07
doc: update the manual deployment on the rke2 cluster

Vicente-Cheng:
related issues: https://github.com/harvester/harvester/issues/2755

PR #295, opened 2023-03-26
point to new script name

qrkourier:


guangbochen:
This will be fixed in another PR https://github.com/harvester/docs/pull/288, but still thanks for your contribution.

PR #329, opened 2023-06-07
Add shadows to images via custom.css

LucasSaintarbor:
### **Summary**
- Added shadows to images via src/css/custom.css. Any CSS included here will be global.
- To avoid setting a shadow to the logo, the style is based on the class ID (img_ev3q) which is set by Docusaurus when building the docs. The class ID is constant across the builds.
- Referencing https://github.com/rancher-sandbox/docs.rancherdesktop.io/pull/140


vickyhella (approved):
LGTM!

PR #242, opened 2022-11-09
[WIP] Add swagger config

vickyhella:
This PR:

- Adds the openapi plugin `2.0.0-beta.3`.
- Bumps docusaurus to `2.4.1`
- Bumps react to `18.2.0`
- Adds swagger API config to the `docusaurus.config.js` file 
- Generates API mdx files

> The `mdx` files in the docs/api folder were generated by running the `yarn docusaurus gen-api-docs all` command.

> Compatibility:
> Docusaurus `2.3.1` works well with openapi `2.0.0-beta.1`.
> Docusaurus `2.2.0` works well with openapi `1.4.2`.
> Docusaurus `2.4.1` doesn't work with openapi `2.0.0-beta.2` or `2.0.0-beta.3`.
> **Update**: after upgrading react to the latest `18.2.0`, docusaurus and openapi can be bumped to the latest `2.4.1` and `2.0.0-beta.3`.

## Related issue: 
- https://github.com/harvester/harvester/issues/3042

## What needs to be done:
- Modify `api/petstore.yaml` and `docusaurus.config.js` based on our actual needs (see the plugin documentation for more info). Note that the current `api/petstore.yaml` file is an example file for reference only.
- Run `yarn docusaurus gen-api-docs all`, which generates MDX files in the `docs/api` folder.
- It'll be better if we can split the yaml file into several files by category, because the plugin actually generates one page for each API, which might be hard to read as we have hundreds of APIs.

  image

## References
- Plugin documentation: https://docusaurus-openapi.tryingpan.dev/
- Official demo: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/tree/main/demo
- Example json/yaml config: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/tree/main/demo/examples
- Example `docusaurus.config.js` config: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/blob/main/demo/docusaurus.config.js

PR #330, opened 2023-06-07
Upgrade Docusaurus from v2.3.1 to v2.4.1

LucasSaintarbor:
### **Summary**
- Docusaurus recommended updating to v2.4.1
- Ran the following recommended `yarn` command: `yarn upgrade @docusaurus/core@latest @docusaurus/plugin-client-redirects@latest @docusaurus/plugin-content-blog@latest @docusaurus/plugin-content-docs@latest @docusaurus/plugin-content-pages@latest @docusaurus/plugin-debug@latest @docusaurus/preset-classic@latest`

**Before**
Screenshot 2023-06-07 at 2 52 47 PM

**After**
Screenshot 2023-06-07 at 2 55 22 PM


vickyhella:
@LucasSaintarbor We have another unmerged PR for the Harvester API docs (#242), which requires Docusaurus v2.3.1.
I have tested if 2.4.1 works for the API docs, and also tried bumping the plugin to the latest version 2.0.0-beta3, but the error still occurs.
Therefore, if the bump is not a must for the time being, maybe we can continue using v2.3.1 for the moment and bump it when the compatibility issue is addressed.

LucasSaintarbor:
> @LucasSaintarbor We have another unmerged PR for the Harvester API docs (#242), which requires Docusaurus v2.3.1. I have tested if 2.4.1 works for the API docs, and also tried bumping the plugin to the latest version 2.0.0-beta3, but the error still occurs. Therefore, if the bump is not a must for the time being, maybe we can continue using v2.3.1 for the moment and bump it when the compatibility issue is addressed.

@vickyhella Got it, thanks for the heads up. Yes, sounds good! I'll keep an eye on #242

vickyhella:
@LucasSaintarbor  After bumping `react` to the latest v18.2, Docusaurus v2.4.1 can now work with the latest openapi plugin version. Therefore, I am merging this PR~ If you have any further suggestions, please let me know.

vickyhella (approved):

Issue #341, opened 2023-06-22
Add canonical links


LucasSaintarbor:
### Summary

In order to improve SEO for Rancher projects, Harvester should implement canonical href links to optimize search querying and to ensure users are directed to the latest documentation.
- Canonical references indicate which version of a page is "canon" – in practical terms, which page should appear at the top of the search results.
- Canonical references are added to our docs' Markdown pages in HTML tags. The link is an URL to the published page on the live site. 

Add canonical links with this script: https://github.com/btat/rm-seo-helpers. The following is added to the top of each page.
```
<head> 
  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/other-installation-methods/air-gapped-helm-cli-install/install-rancher-ha"/>
</head>
```

### Next

Open PR to add canonical links by folder
- [x] volume
- [x] vm
- [x] upgrade
- [x] troubleshooting
- [x] terraform
- [x] rancher
- [x] networking
- [x] monitoring
- [x] logging
- [x] install
- [x] host
- [x] advance
- [x] files [airgap.md, authentication.md, faq.md, index.md, upload-image.md]

LucasSaintarbor:
Adding canonical links for v1.2 via https://github.com/harvester/docs/issues/446

PR #350, opened 2023-07-12
Add canonical links /vm 

LucasSaintarbor:
Adding canonical links to /vm section pages in Harvester.

Tied to #341

PR #359, opened 2023-07-14
Add canonical links for /volume

LucasSaintarbor:
Adding canonical links to /volume section pages in Harvester.

Tied to https://github.com/harvester/docs/issues/341


vickyhella:
Hi Lucas, I think #350 already adds links for the `volume` section. In this case, please decide if we should close this volume-specific PR, or remove the volume parts from #350 ^^

vickyhella (approved):

PR #370, opened 2023-07-27
Fix image positions for the volume section

vickyhella:
Put step-related images right behind the corresponding step to avoid piling up the images.

### DIFF
--- a/docs/volume/clone-volume.md
+++ b/docs/volume/clone-volume.md
@@ -14,10 +14,11 @@ Description: Clone volume from the Volume page.
 After creating a volume, you can clone the volume by following the steps below:
 
 1. Click the `⋮` button and select the `Clone` option.
-1. Select `clone volume data`
+
+    ![clone-volume-1](/img/v1.2/volume/clone-volume-1.png)
+
+1. Select `clone volume data`.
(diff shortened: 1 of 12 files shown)


LucasSaintarbor:
Hi @vickyhella, thanks for updating these! I noticed some of the images were wrapped with the sentence above for the following pages:

- **Export a Volume to Image** > English v1.1 and v1.2-dev
Screenshot 2023-07-27 at 11 14 46 AM


- **Volume Snapshots** > Chines v1.1 and v1.2-dev
Screenshot 2023-07-27 at 11 19 19 AM


- I think we should add a space between the step and the top of the image, as well as the after the image so it doesn't look cluttered. Do you agree?
Screenshot 2023-07-27 at 11 21 34 AM

LucasSaintarbor (changes requested):
Added feedback here: https://github.com/harvester/docs/pull/370#issuecomment-1654183081

vickyhella:
@LucasSaintarbor Thanks for noticing it! 
I've updated this PR using empty lines to separate images and text. Please take a second look and thanks ^^

LucasSaintarbor (approved):
LGTM, thanks @vickyhella!

PR #383, opened 2023-08-09
doc: Update Harvester Addon Config Info

irishgordo:
Super willing to iterate and update this as needed but wanted to push up a slight change :smile: 

Given that there is a distinction between UI "naming" and Helm Chart naming:
Screenshot from 2023-07-26 11-36-32
Screenshot from 2023-07-26 11-32-52

And that in order for the configuration to work, if serving remotely as a configuration, the syntax in the YAML will need to change shifting from `-` (hypens) to `_` (underscores).

Additionally, sharing out to docs about `values_content` as a property on the addon.

This PR updates the way the user is informed of:
- proper YAML configuration for enablement
- the distinction between "chart name"/"chartname" (helm chart name) and what the user would see through the Web UI (if enabling through WebUI/harvester-dashboard, instead of kubectl edits)
- additionally providing the `valuescontent` property as a configuration element 

Newer PR from: https://github.com/harvester/docs/pull/368

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -75,10 +75,21 @@ install:
   vip_mode: dhcp
   force_mbr: false
   addons:
-    rancher-monitoring:
-      enabled: true
-    rancher-logging:
+    harvester_vm_import_controller:
       enabled: false
+      values_content: ""
+    harvester_pcidevices_controller:
+      enabled: false
+      values_content: ""
+    rancher_monitoring:
+      enabled: true
+      values_content: ""
+    rancher_logging:
+      enabled: false
+      values_content: ""
+    harvester_seeder:
+      enabled: false
+      values_content: ""
 system_settings:
   auto-disk-provision-paths: ""
 ```
@@ -293,7 +304,7 @@ The password for the default user, `rancher`. By default, there is no password f
 If you set a password at runtime it will be reset on the next boot. The
 value of the password can be clear text or an encrypted form. The easiest way to get this encrypted
 form is to change your password on a Linux system and copy the value of the second field from
-`/etc/shadow`. You can also encrypt a password using OpenSSL. For the encryption algorithms 
+`/etc/shadow`. You can also encrypt a password using OpenSSL. For the encryption algorithms
 supported by Harvester, please refer to the table below.
 
 | Algorithm | Command | Support |
@@ -566,18 +577,19 @@ Default: The addons are disabled.
 ```yaml
 install:
   addons:
-    rancher-monitoring:
+    rancher_monitoring:
       enabled: true
-    rancher-logging:
+    rancher_logging:
       enabled: false
 ```
 
-Harvester v1.2.0 ships with four addons:
+Harvester v1.2.0 ships with five addons:
 
-- harvester-vm-import-controller
-- harvester-pcidevices-controller
+- vm-import-controller (chartName: `harvester-vm-import-controller`)
+- pcidevices-controller (chartName: `harvester-pcidevices-controller`)
 - rancher-monitoring
 - rancher-logging
+- harvester-seeder (experimental)
 
 ### `install.harvester.storage_class.replica_count`
 


guangbochen (approved):
LGTM, thanks.

docs/install/harvester-configuration.md
```diff
@@ -574,8 +582,8 @@ install:
 
 Harvester v1.2.0 ships with four addons:
 
-- harvester-vm-import-controller
-- harvester-pcidevices-controller
+- vm-import-controller (chartName: `harvester-vm-import-controller`)
+- pcidevices-controller (chartName: `harvester-pcidevices-controller`)
 - rancher-monitoring
 - rancher-logging
```
bk201, line 588: We have a new one rancher-vcluster (experimental)

docs/install/harvester-configuration.md
```diff
@@ -75,10 +75,18 @@ install:
   vip_mode: dhcp
   force_mbr: false
   addons:
-    rancher-monitoring:
+    harvester_vm_import_controller:
```
bk201, line 78: @w13915984028 Please help check this part. Thanks!

w13915984028:
@irishgordo:

Per the past test, when put `rancher-monitoring` in the `harvester-configuration` to enable the addon by default, the `rancher-monitoring` works, I did not test `rancher_monitoring`, did you tried it?

IIRC, string like `rancher_monitoring` in the YAML will be converted to `rancherMonitoring`.

docs/install/harvester-configuration.md
w13915984028: need to double check:

IIRC, string like rancher_monitoring in the YAML will be converted to rancherMonitoring in bootstrap stage.

w13915984028:
The string like `a_b` in the YAML file, will be converted to `aB` (camelCase) in the bootstrap stage;  in fact, even `aB` in the YAML file can also work. Harvester follows the rule from Rancher, claims to use `a_b` to represent `aB` to Harvester configuration file.

For `addon` name, I need to check more, because `addon` name is in format `a-b`, it is not targeted as `aB`.

docs/install/harvester-configuration.md
w13915984028: there is also a seeder addon:

https://github.com/harvester/harvester-installer/blob/master/pkg/config/templates/rancherd-22-addons.yaml
```
  - apiVersion: harvesterhci.io/v1beta1
    kind: Addon
    metadata:
      name: harvester-seeder
      namespace: harvester-system
    labels:
      addon.harvesterhci.io/experimental: "true"
    spec:
  ```

Comment (w13915984028) [reply docs/install/harvester-configuration.md:588]: The `chartName` comes for the first time in this file, we may add an explanation of `chartname` here, then say it is same as addon name by default, for those who have different names, they are listed separately.

Comment (w13915984028) [reply docs/install/harvester-configuration.md:588]: For `vcluster-rancher`, @ibrokethecloud mentioned, it is not packed into Harvester OS, namely, not possible to be enabled via the `harvester-configuration`.

Need to include it in the coming `vcluster-rancher` chapter in the Harvester Doc.

**Comment by irishgordo** [comment]
> @irishgordo:
> 
> Per the past test, when put `rancher-monitoring` in the `harvester-configuration` to enable the addon by default, the `rancher-monitoring` works, I did not test `rancher_monitoring`, did you tried it?
> 
> IIRC, string like `rancher_monitoring` in the YAML will be converted to `rancherMonitoring`.

@w13915984028  thanks for that mention :smile:  - I did try the `rancher_monitoring` w/ success per my:
https://github.com/irishgordo/scratch-dump/blob/main/test-vm.yaml

Remote Harvester Configuration on Interactive ISO Install ^

And additionally tested w/ R720:
- https://github.com/irishgordo/scratch-dump/blob/main/dell-poweredge-r720-create.yaml

Screenshot from 2023-07-26 11-32-52

I wasn't able to get:
```yaml
    rancher-monitoring:
      enabled: true
      valuescontent: ""
```
To work on rc4, if I recall correctly (but perhaps that was bad configuration on my end :thinking: ), hence the change to the `_` underscore

docs/install/harvester-configuration.md
irishgordo: great mention :smile: :+1: - I've updated this in the latest force push  :smile:

irishgordo:
@w13915984028 & @ibrokethecloud 
I've updated the configuration to take into account:
- harvester-seeder
- rancher-vcluster

So we then with v1.2.0, will be shipping with six addons correct?

w13915984028:
@irishgordo Sorry I confused you, the harvester-configuration do need to use "rancher_monitoring" to express the addon name.

For addons in Harvester v1.2.0, from source code https://github.com/harvester/harvester-installer/blob/master/pkg/config/templates/rancherd-22-addons.yaml,

The `harvester-seeder` has `experimental` label;  and the `vcluster-rancher` is not in the list, it means user can't enable it from the harvester-configuration. @ibrokethecloud mentioned the `vcluster-rancher` is a special addon and not shipped within the ISO. In this commit, we can skip `vcluster-rancher` addon. thanks.

```
  - apiVersion: harvesterhci.io/v1beta1
    kind: Addon
    metadata:
      name: harvester-seeder
      namespace: harvester-system
    labels:
      addon.harvesterhci.io/experimental: "true"
```

irishgordo:
@w13915984028 oh okay :smile: - that sounds good to me :+1:  - I've pushed up a forced amended commit update that adds the experimental note to `harvester-seeder` and removed the rancher-vcluster

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
lgtm!

LucasSaintarbor (approved):

Issue #307, opened 2023-05-04
Revise Harvester documentation for formatting, style, and grammar


LucasSaintarbor:
## Description
Revise and improve Harvester docs for formatting, style, and grammar.

## Goal
Normalize content structure and styling. Each page is revised against the following checklist:

- Revise page locally (https://github.com/LucasSaintarbor/harvester-docs)
- Content adheres to [Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/top-10-tips-style-voice)
- Headers are accurate
- Lists are formatted properly (ordered vs unordered)
- Images are clear and sized appropriately
- Validate links work and are accurate
- Open PR
- Ask for review from docs team member
- Close PR

- Reference [Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/top-10-tips-style-voice) and [Divio Documentation System](https://documentation.divio.com/). 

## Pages
Update pages once reviewed by engineering per https://github.com/harvester/docs/issues/373

- [x] Harvester Overview
- [ ] Installation
  - Hardware and network requirements: https://github.com/harvester/docs/pull/377
  - ISO Installation: https://github.com/harvester/docs/pull/396
  - USB Installation: https://github.com/harvester/docs/pull/403
  - PXE Boot Installation
  - Harvester Configuration
  - Update Harvester Configuration
  - Management Address

- [ ] Air Gapped Environment
- [ ] Upgrade
- [ ] Authentication
- [ ] Upload Images
- [ ] Host Management
- [ ] VM Management
- [ ] Volume Management
- [ ] Networking
- [ ] Logging
- [ ] Monitoring
- [ ] Advanced
- [ ] Rancher Integration
- [ ] Harvester Terraform Provider
- [ ] Troubleshooting
- [ ] FAQ

PR #412, opened 2023-09-01
Add images and update formatting for ISO Installation page

LucasSaintarbor:
- Added images and updated formatting so v1.1 and v1.2 are aligned (https://github.com/harvester/docs/issues/307).


vickyhella (approved):
Only a small nit and the rest is good to me!

PR #388, opened 2023-08-15
Add swagger config and API docs

vickyhella:
## Summary

- Adds the openapi plugin `2.0.0-beta.3`
- Bumps react to 18.2.0 to avoid plugin incompatibilities
- Adds swagger config to the `docusaurus.config.js` file
- Generates API mdx files for v1.2 and v1.1 (run `yarn docusaurus clean-api-docs all && yarn docusaurus gen-api-docs all && yarn docusaurus clean-api-docs:version api:all && yarn docusaurus gen-api-docs:version api:all`)
- Uses autogenerated sidebars for both docs and API pages

## Related issues and PRs

This PR addresses https://github.com/harvester/harvester/issues/3042 and #373, and replaces #242

## Screenshots

> Clicking the version switch on the API pages will switch to the API pages of the specified version (v1.1 and v1.2 only).

image

## References

- Plugin documentation: [docusaurus-openapi.tryingpan.dev](https://docusaurus-openapi.tryingpan.dev/)
- Official demo: [PaloAltoNetworks/docusaurus-openapi-docs@main/demo](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/tree/main/demo?rgh-link-date=2022-11-09T09%3A31%3A28Z)
- Example json/yaml config: [PaloAltoNetworks/docusaurus-openapi-docs@main/demo/examples](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/tree/main/demo/examples?rgh-link-date=2022-11-09T09%3A31%3A28Z)
- Example docusaurus.config.js config: [PaloAltoNetworks/docusaurus-openapi-docs@main/demo/docusaurus.config.js](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/blob/main/demo/docusaurus.config.js?rgh-link-date=2022-11-09T09%3A31%3A28Z)

PR #416, opened 2023-09-04
Update broken link settings for CI to pass

### LINKED ISSUE #415: Docusaurus broken link settings cause build failure for English docs
## Issue Description

As the English and Chinese docs are not developed and submitted at the same time, when an English page is added, sometimes CI may fail for the zh locale because the Chinese doc doesn't have the corresponding content when the PR is created.

For the build failure details, see https://drone-pr.rancher.io/harvester/docs/143/1/4

### Screenshot

image

## Solution

Therefore, we need to change the security settings `onBrokenLinks` to a lower one `warn` (default) for our CI to pass during doc development.

## Reference

https://docusaurus.io/docs/api/docusaurus-config#onBrokenLinks

vickyhella:
This PR:

1. Fixes broken links found in the existing docs.
2. Changed `dev` to `v1.2` in v1.2-related links.
3. Adds back the `remark-validate-links` plugin that was accidentally deleted in #388.
4. Sets `onBrokenLinks` to `warn` to avoid CI failure caused by the update time gap between the CN and EN docs during documentation development.
5. Use relative links rather than URLs.
6. Tweaks the remark config to better facilitate local link checks.

## Related issues

Fixes #415

### DIFF
--- a/docs/advanced/rancher-vcluster.md
+++ b/docs/advanced/rancher-vcluster.md
@@ -5,7 +5,7 @@ title: "Rancher vcluster (Experimental)"
 ---
 
 <head>
-  <link rel="canonical" href="https://docs.harvesterhci.io/dev/advanced/rancher-vcluster"/>
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.2/advanced/rancher-vcluster"/>
 </head>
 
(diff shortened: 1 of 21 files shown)


guangbochen (approved):
LGTM, thanks.

PR #421, opened 2023-09-06
Use "v1.2" for v1.2 doc URLs

vickyhella:
Use `v1.2` in v1.2 doc URLs instead of `dev` so that we can use the correct URLs in the Release Notes and the Dashboard.


LucasSaintarbor (approved):
Thanks @vickyhella!

PR #426, opened 2023-09-08
Convert URLs to relative paths

vickyhella:
Converts URLs to relative paths so that the links can point to the right version of docs

We should avoid using URLs in the docs:
- The URLs were added when we developed the v1.1 docs, and the docs were then reused in v1.2. These URLs pointing to v1.1 do not apply to v1.2 docs.
- Using relative paths makes sure that the links always point to the corresponding doc of the desired version.


LucasSaintarbor (approved):
LGTM, thanks @vickyhella!

PR #434, opened 2023-09-14
Update Chinese docs and remove doc IDs

### LINKED ISSUE #431: Need to remove the doc id from markdown files
## Background
We added doc IDs to v1.2 and v1.1 docs to use manual sidebars for the API reference docs in #388. 
However, we later found a way to use autogenerated sidebars instead. Therefore, the doc IDs can now be safely removed to reduce complexity.

## Benifits
After the IDs are removed, doc writers won't have to be confused about whether to use IDs or what IDs to use.

vickyhella:
- Updates Chinese translation for v1.2, v1.1, and v1.0
- Remove IDs from markdown files (fixes #431)


LucasSaintarbor (approved):
Reviewed /docs and /version_docs. LGTM, thanks!

PR #459, opened 2023-09-20
Fixed typo & grammar in rancher-integration.md

simonflood:
abled > able
in a different provider > with a different provider


LucasSaintarbor (approved):
@simonflood thanks for the fix!

PR #491, opened 2023-12-15
Enable dark mode option

jillian-maroket:
This change adds a button to the rightmost area of the header. The button allows users to switch from light mode (default) to dark mode with one click.


LucasSaintarbor (approved):
LGTM 👍 Thanks @jillian-maroket!

PR #498, opened 2024-01-12
doc: update Port Requirements

Vicente-Cheng:
- Now, we do not expose the NDM profiler port

Related issue: https://github.com/harvester/harvester/issues/4212


bk201:
will this be backport to v1.2.x?

Vicente-Cheng:
> will this be backport to v1.2.x?

I will update to v1.2 once it is backported to v1.2.x
The 1.2.1 still exposes the 6060 port, so we can keep it on the v1.2 document now.

jillian-maroket (approved):

PR #506, opened 2024-01-30
Updating Data Locality for Storage Class and Virtual Machine Image

WebberHuang1118:


bk201 (approved):

docs/advanced/storageclass.md
```diff
@@ -97,6 +97,29 @@ The `volumeBindingMode` field controls when volume binding and dynamic provision
 
 ![](/img/v1.2/storageclass/customize_tab_vol_binding_mode.png)
 
+## Data Locality Settings
+
+The `data locality` setting is intended to be enabled in situations where at least one replica of a Longhorn volume should be scheduled on the same node as the pod that uses the volume, whenever it is possible.
+
+Harvester officially supports `data locality` since **v1.3.0**, especially for the volumes created from the [Images](../upload-image.md), such volumes didn't support `data locality` before. Users could configure the `data locality` from **Storage Classes > Create > Parameters > Add Parameter**, with **Key** as `dataLocality`, and **Value** as `disabled` or `best-effort`.
```
jillian-maroket, line 104:
```suggestion
You can use the `dataLocality` parameter when at least one replica of a Longhorn volume must be scheduled on the same node as the pod that uses the volume (whenever possible).

Harvester officially supports data locality as of **v1.3.0**. This applies even to volumes created from [images](../upload-image.md). To configure data locality, create a new StorageClass on the Harvester UI (**Storage Classess** > **Create** > **Parameters**) and then add the following parameter:

- **Key**: `dataLocality`
- **Value**: `disabled` or `best-effort`
```

docs/advanced/storageclass.md
```diff
@@ -97,6 +97,29 @@ The `volumeBindingMode` field controls when volume binding and dynamic provision
...
+### best-effort
+
+This instructs Longhorn to try its best to keep a replica on the same node for the attached volume. Longhorn will not stop the volume, even if it cannot keep a replica local due to an environment limitation, e.g. not enough disk space, incompatible disk tags, etc.
+
+:::note
+
+- You may refer to [Longorn document](https://longhorn.io/docs/1.6.0/high-availability/data-locality/) for more details.
+- Longhorn supports another `data-locality` setting as **strict-local**, however, Harvester prevents such setting, since this enforces the volume could has **only one replica**. It will brings some limitations, like [VM Live Migration](../vm/live-migration.md) .. and so on.
+
+:::
```
jillian-maroket, line 121:
```suggestion
### Data Locality Options

Harvester currently supports the following options:

- `disabled`: When applied, Longhorn may or may not schedule a replica on the same node as the pod that uses the volume. This is the default option. 

- `best-effort`: When applied, Longhorn always attempts to schedule a replica on the same node as the pod that uses the volume. Longhorn does not stop the volume even when a local replica is unavailable because of an environmental limitation (for example, insufficient disk space or incompatible disk tags).

:::note
Longhorn provides a third option called `strict-local`, which forces Longhorn to keep only one replica on the same node as the pod that uses the volume. Harvester does not support this option because it can affect certain operations such as [VM Live Migration](../vm/live-migration.md)
:::

For more information, see [Data Locality](https://longhorn.io/docs/1.6.0/high-availability/data-locality/) in the Longhorn documentation.
```

jillian-maroket: I used a list because the options do not require their own headings.

jillian-maroket (commented):
@WebberHuang1118 Let me know if you have concerns about the changes.

WebberHuang1118:
Hi @jillian-maroket,
Updated per suggestions, thanks.

jillian-maroket (approved):

PR #513, opened 2024-02-21
Adding instruction to upload image from multi-cluster UI

WebberHuang1118:


jillian-maroket (commented):
@WebberHuang1118 Let me know if you have concerns about the changes.

WebberHuang1118:
Hi @jillian-maroket,

Updated per suggestion, and sync the modification with the v1.2 version. 
Besides, I also had a slight update on the paragraph `for example, 1 MB ... K3S clusters`, please help to check it, thanks.

jillian-maroket (approved):
> (for example, 1 MB is the default limit in RKE2 clusters, however, no limit is defined in K3S clusters)

LGTM overall. Just please change this to "(for example, RKE2 clusters have a default limit of 1 MB but no such limit exists in K3S clusters)". Thanks!

PR #514, opened 2024-02-27
Fix list indent on Custom SUSE VM Images page

tserong:
By indenting the screenshots and other text, we can go back to using "1." everywhere as the numbered list item prefix, so don't have to worry about counting list items (e.g. the explicit "3." I added earlier is no longer necessary).  Also, this makes the screenshots indent nicely when they appear inside list items.

The only thing I'm not sure about is whether it matters that the following screenshots under the later headings ("3. (Optional) Switch to View Package mode", etc.) are not indented the same as the ones that appear inside lists?

### DIFF
--- a/docs/advanced/customsuseimages.md
+++ b/docs/advanced/customsuseimages.md
@@ -32,29 +32,29 @@ You can create custom images based on what SUSE provides using OBS [image templa
 
 1. Specify a name for the image, and then select **Create appliance**.
 
-![](/img/v1.3/advanced/custom-vm-01-select-template.png)
+   ![](/img/v1.3/advanced/custom-vm-01-select-template.png)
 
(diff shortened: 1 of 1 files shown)


tserong:
A small visual tweak - wait 'til the preview is built, then compare indenting of screenshots with https://docs.harvesterhci.io/v1.3/advanced/customsuseimages/ and you'll see what I mean.

tserong:
Preview link to compare is https://65dda1dc38c934a2ed5ea136--harvester-preview.netlify.app/v1.3/advanced/customsuseimages

jillian-maroket:
@tserong Thanks for catching the indentation issue. I think we should allow the content structure to dictate when objects are indented. The latter sections, which have simpler structures, don't have to align with the ones with lists.

PR #517, opened 2024-03-04
Add workaround for download image fail during storage network is configured

WebberHuang1118:


bk201 (approved):
lgtm.

To save user from typing, sometimes it's OK to export the image name in question into a environment variable and re-use that environment variable in the commands.

WebberHuang1118:
@jillian-maroket,
Updated per suggestion, thanks.

jillian-maroket (approved):
LGTM. Thanks!

PR #524, opened 2024-03-12
Add temporary search via @cmfcmf/docusaurus-search-local

LucasSaintarbor:
Issue: 
- A 10000 search limit in Algolia was hit for the first time, and searches are trending upward. It may be worth upgrading to the Algolia Grow Plan. 
- The quota will refresh on 03/14. 

image
image

Summary: 
- Currently testing a local search option similar to (https://github.com/cmfcmf/docusaurus-search-local) as a temporary solution.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -70,6 +70,7 @@ const config = {
   ],
   themeConfig: {
     zoom: {},
+    /*
     algolia: {
       appId: 'U7QCSJFCWR',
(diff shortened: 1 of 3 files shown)


jillian-maroket (approved):
Thank you for testing and implementing this workaround.

jillian-maroket:
As discussed, we will use the [docusaurus-search-local](https://github.com/cmfcmf/docusaurus-search-local) plugin while waiting for the Algolia counter to reset.

PR #537, opened 2024-03-27
Document management address (VIP) requirements

connorkuehl:
**Problem:**

The requirements for the Harvester management address VIP are not well-defined in the docs

**Solution:**

Add some requirements and also a warning about how some bare metal providers don't technically meet them and how it can cause issues

**Related Issue:** https://github.com/harvester/harvester/issues/5417

### DIFF
--- a/docs/install/management-address.md
+++ b/docs/install/management-address.md
@@ -21,6 +21,28 @@ If you selected the IP address to be configured via DHCP, you will need to confi
 
 ![](/img/v1.2/install/iso-installed.png)
 
+## Requirements
+
+1. The VIP must belong to the same subnet as the node management interfaces.
+1. All of the Harvester node management interfaces must be on the same layer-2 network segment.
+
+Both of these requirements are because the VIP relies on the Address Resolution Protocol (ARP), which is a layer-2 protocol.
+The VIP can be assigned to any one of the Harvester management interfaces, and it can change at any time, not just when there is a node failure.
+
+When the VIP changes hosts, it sends out "gratuitous ARPs" so that the rest of the hosts on the network know where to direct traffic that is meant for the Harvester cluster's management IP address (the VIP.)
+
+:::warning
+
+If you are planning to host a Harvester cluster in a bare metal provider's data center, it is possible that your provider will give you a "floating" or "reserved" IP address to act as the VIP that can be assigned to one of your servers at your discretion. Updating this assignment causes traffic to instantly start flowing to the new node it is assigned to.
+
+In these situations, Harvester is unable to update that floating/reserved IP assignment in your bare metal provider's systems when the VIP changes hosts.
+
+Furthermore, the VIP's "gratuitous ARPs" may also be ineffective depending on your provider's networking setup between your Harvester nodes (e.g., if your Harvester hosts are not on the same layer-2 network.)
+
+You will have to update the floating/reserved IP assignment manually when the VIP changes hosts, otherwise the cluster may not function properly. (For more info, see ["Finding which node the VIP is on."](#finding-which-node-the-vip-is-on))
+
+:::
+
 ## How to get the VIP MAC address
 
 To get the VIP MAC address, you can run the following command on the management node:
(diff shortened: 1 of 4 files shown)


docs/install/management-address.md
```diff
@@ -33,6 +55,34 @@ Example of output:
...
+
+```console
+kubectl get pod -n harvester-system -o custom-columns='NAME:.metadata.name,NODE:.spec.nodeName' | grep -E '^kube-vip|^NAME'
+```
+
+Example output:
+```console
+NAME                                                    NODE
+kube-vip-hxzfk                                          harvester-xzj76
+```
```
starbops, line 70: This part might need to be more precise. The `kube-vip` Pods run on control-plane nodes, so if there are three Harvester management nodes, three Pods will be listed in the command output, and people still don't know which node the VIP address is on.

I'd suggest we could use the following instead:

```shell
kubectl -n kube-system get svc ingress-expose -o jsonpath='{.metadata.annotations.kube-vip\.io/vipHost}'
```

starbops (commented):
Thank you for the doc enhancement. This is really helpful, and it also addresses harvester/harvester#5324.

About the floating IP section, I'm unsure if that's generally the case because in my experience in OpenStack (sorry, I don't have public cloud experiences, so please correct me if I'm wrong), the floating IPs are usually within an entirely different subnet, which is separated from the one VMs attach to. Then, the incoming traffic toward the floating IP will be DNAT'd to a mapped internal IP, which should be the Harvester VIP in our case. So I think it should be fine in terms of floating IP.

connorkuehl:
> About the floating IP section, I'm unsure if that's generally the case because in my experience in OpenStack (sorry, I don't have public cloud experiences, so please correct me if I'm wrong), the floating IPs are usually within an entirely different subnet, which is separated from the one VMs attach to. Then, the incoming traffic toward the floating IP will be DNAT'd to a mapped internal IP, which should be the Harvester VIP in our case.

Yeah, that would be one way to do it. However, what prompted me filing this pull request was Equinix Metal requires kube-vip to speak BGP in order to update their software defined networking, otherwise Equinix Metal will _always_ route the packet to the machine that the floating IP is assigned to. The fix for this particular issue was to manually update what machine the floating IP was pointing to in Equinix Metal.

docs/install/management-address.md
```diff
@@ -33,6 +55,33 @@ Example of output:
 {"kube-vip.io/hwaddr":"02:00:00:09:7f:3f","kube-vip.io/requestedIP":"10.84.102.31"}
 ```
 
+## Finding which node the VIP is on
+
+Using `kubectl` (either on your own machine with the Harvester kubeconfig, or SSH'd into any Harvester management node as root):
```
jillian-maroket, line 60:
```suggestion
## Identifying the Node the VIP Is Assigned To

You can use kubectl (either on your local machine with the Harvester kubeconfig file, or when using SSH to connect to any Harvester management node as a root user).
```

docs/install/management-address.md
```diff
@@ -33,6 +55,33 @@ Example of output:
...
+```console
+kubectl -n kube-system get svc ingress-expose -o jsonpath='{.metadata.annotations.kube-vip\.io/vipHost}'
+```
+
+Example output:
+```console
+harvester-xzj76
+```
+
+Alternatively, you can SSH into each Harvester management node and run:
```
jillian-maroket, line 71:
```suggestion
Alternatively, you can use SSH to connect to each Harvester management node and then run the command `ip address show mgmt-br`.

Example:
```

docs/install/management-address.md
```diff
@@ -33,6 +55,33 @@ Example of output:
...
+harvester-xzj76:~ # ip address show mgmt-br
+4: mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
+    link/ether 8c:dc:d4:b5:f0:fd brd ff:ff:ff:ff:ff:ff
+    inet 172.19.108.45/21 brd 172.19.111.255 scope global mgmt-br
+       valid_lft forever preferred_lft forever
+    inet 172.19.108.34/32 scope global mgmt-br
+       valid_lft forever preferred_lft forever
+```
+
+When you see both the VIP and the node address in the output of that command, then that means that the node you are running on is the one with the VIP attached to it.
```
jillian-maroket, line 83:
```suggestion
When the output includes both the VIP and the node address, the VIP is assigned to the node that you are connected to.
```

jillian-maroket (commented):

connorkuehl:
Thanks, @jillian-maroket. I'll update it to your version once I hear back from @starbops regarding some of the details.

starbops (approved):
@connorkuehl Thanks for the kind explanation regarding how Equinix Metal provides VIP services. LGTM!

jillian-maroket (approved):
LGTM. Thanks!

PR #544, opened 2024-04-09
Removing extra formatting

asettle:
The formatting on the vGPU page is broken for the notation. See: https://docs.harvesterhci.io/v1.3/advanced/vgpusupport


jillian-maroket:
I created [PR 556](https://github.com/harvester/docs/pull/556), which applies the fix to both v1.3 and v1.4 docs.

PR #558, opened 2024-05-07
ci: fix release build

m-ildefons:
Fix the documentation's release build process by adding a step, which generates the API documentation, prior to building the entire documentation page and publishing it.
This is necessary because PR#545 removed the pre-generated API documentation from the source repository. As a result, prior to building the documentation, the API documentation now needs to be re-generated.


jillian-maroket (approved):
Thank you for getting this done quickly.

PR #545, opened 2024-04-10
API update, build fixes, add some API flows

m-ildefons:
- API update:
Add a v1.3 API reference source and a source for the current development version.

- Build fixes:
Upgrade the openapi-docs plugin for Docusaurus to a stable version
Downgrade Docusaurus to 2.4.0, since 2.4.1 is incompatible with the openapi-docs plugin. This allows us to generate API reference docs from swagger.json files. These swagger.json files can be generated from the Golang type definitions of our CRDs in the harvester/harvester repository
The readme is also updated to add instructions how to maintain the API docs.

- Remove generated content
Existing generated content may confuse the openapi-docs plugin for docusaurus. To make sure that a clean git repository will also always be a clean basis for modifying the documentation, including the API reference docs, the generated files are removed. They can easily be re-generated if needed.

- Add some API flows.
Add documented API flows for uploading a VM image, creating a VM and registering a Harvester cluster in a Rancher Manager instance.
There are still many flows missing, but these three should serve as good examples, how our documentation could include both UI based flows and API based flows in the future:

https://github.com/harvester/docs/assets/17141774/2e56501c-cf9e-4eb5-bdad-235fdbf3c391

PR #564, opened 2024-05-14
Remove docusaurus-lunr-search plugin (to force usage of Algolia)

jillian-maroket:
The Harvester doc currently uses the Docusaurus Lunr plugin instead of Algolia, which should NOT be the case. PR #545 (API update, build fixes, add some API flows) added back the Lunr plugin that we removed in PR #547 (Reactivate Algolia search). #547 was merged before #545. 

Screenshot 2024-05-14 at 2 39 06 PM

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -233,15 +233,6 @@ const config = {
   plugins: [
     require.resolve('docusaurus-plugin-image-zoom'),
     [
-      require.resolve('docusaurus-lunr-search'), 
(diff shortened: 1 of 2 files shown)


m-ildefons (approved):
LGTM.
Sorry for adding this back. I guess it happened during rebasing.

bk201:
Thank you @jillian-maroket and @m-ildefons , Algolia is back now.

PR #568, opened 2024-05-20
docs: update swaggers api

Yu-Jack:
https://github.com/harvester/harvester/issues/5497


bk201 (approved):
lgtm

Yu-Jack:
FYI, I cherry-pick new generator commit on v1.3.0 and master to generate two open api files. So, this PR modified dev and v1.3.0 open api.

jillian-maroket (approved):
@Yu-Jack and @bk201 Sorry for the delay. I have no concerns about the changes in this PR, but I suggest changing the way we write the summary and description text. This will require a separate PR.

The general idea is to provide more useful information in the description (not just repeat the summary). Here are some examples of possible changes:

"read" requests
summary: Get %OBJECT-TYPE% object details 
definition: Displays information about the specified %OBJECT-TYPE% object.

"list" requests
summary: List %OBJECT-TYPE% objects
description: Displays information about all %OBJECT-TYPE% objects that match the specified criteria (in a paginated list?).

"patch" requests
summary: Patch a %OBJECT-TYPE% object
description: Modifies the specified %OBJECT-TYPE% object.

Yu-Jack:
@jillian-maroket Thank you for suggestions. That json files are automatically generated right now. So, it would be hard to add specified summary and explanation. But, it's still valuable as a reference.

PR #575, opened 2024-06-04
Versioned Sidebars: Add API sidebar links

m-ildefons:
Add API link definitions for the versioned sidebars for v1.1 and v1.2 API docs. This is used by docusaurus to generate the landing page for the v1.2 and v1.1 API docs. Having this landing page is necessary to be able to navigate to the individual pages of the v1.1 and v1.2 API docs from the regular versioned docs.

fixes: https://github.com/harvester/harvester/issues/5948


Yu-Jack (approved):
LGTM, thanks!

jillian-maroket (approved):
LGTM. I'll create a separate PR for improvements to the top-level title and description, as well as the category labels. Thanks to this PR, I now know which fields to change and where to find them. Thanks again!

PR #576, opened 2024-06-04
ci: preview API docs

m-ildefons:
Generate API docs during preview build. This allows changes to the API docs to be evaluated before deployment to production.


jillian-maroket (approved):
Thank you!

PR #579, opened 2024-06-06
ci: migrate drone to github action

Yu-Jack:
https://github.com/harvester/harvester/issues/5327

test repo - gh-pages branch: https://github.com/Yu-Jack/docs/tree/gh-pages
docs repo gh-pages branch: https://github.com/harvester/docs/tree/gh-pages


Yu-Jack:
Need to merge first, then test testify.

Github pages doesn't need other env, so we could test it right after merged.

bk201 (approved):
LGTM, thanks!

PR #581, opened 2024-06-07
docs: add API and Terraform examples

m-ildefons:
Add example manifests for resources in both the Kubernetes API variant as well as the Terraform resource block variant.

These examples help users to create and manage resources through automatic means by providing example manifests.

The docusaurus config received an update enabling syntax highlighting for Hashicorp configuration languauge (hcl), which is used by Terraform.

Screenshot at 2024-06-07 16-00-40

### DIFF
--- a/docs/terraform/terraform-provider.md
+++ b/docs/terraform/terraform-provider.md
@@ -49,3 +49,28 @@ provider "harvester" {
 More details about the provider-specific configurations can be found in the [docs](https://registry.terraform.io/providers/harvester/harvester/latest/docs).
 
 Github Repo: [https://github.com/harvester/terraform-provider-harvester](https://github.com/harvester/terraform-provider-harvester)
+
+### Resource Timeouts
+
+Several resource-related operations (for example, creating a new image and
+downloading its content from the internet) may take some time to complete.
+Depending on the host hardware and other factors, these operations may exceed
+default timeout settings and cause errors. To modify timeout values for such
+operations, define a timeout block in the resource.
+
+```hcl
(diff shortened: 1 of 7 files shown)


docs/terraform/terraform-provider.md
```diff
@@ -49,3 +49,27 @@ provider "harvester" {
 More details about the provider-specific configurations can be found in the [docs](https://registry.terraform.io/providers/harvester/harvester/latest/docs).
 
 Github Repo: [https://github.com/harvester/terraform-provider-harvester](https://github.com/harvester/terraform-provider-harvester)
+
+### Resource Timeouts
+
+There are several operations on resources that may potentially take some time to complete,
+e.g. creating a new image and downloading its content from the web. Depending on the hardware
+of the hosts and other factors, these operations may exceed default timeout settings, resulting
+in errors. To adjust timeouts of operations on resources, define a timeout block in the resource:
```
jillian-maroket, line 58:
```suggestion
Several resource-related operations (for example, creating a new image and downloading its content from the internet) may take some time to complete. Depending on the host hardware and other factors, these operations may exceed default timeout settings and cause errors. To modify timeout values for such operations, define a timeout block in the resource.
```

jillian-maroket (commented):

docs/terraform/terraform-provider.md
m-ildefons: Thanks for the suggestion, it's much better.
I've updated the PR.

jillian-maroket (approved):
LGTM. Thanks!

PR #611, opened 2024-07-17
doc(networking): add detailed explanation regarding route connectivity

starbops:


docs/networking/harvester-network.md
```diff
@@ -106,3 +106,32 @@ To create a new untagged network, go to the **Networks > VM Networks** page and
...
+
+To check the connectivity, the gateway IP address is of interest to the Harvester network controller. Such information could be absent during  VM Network creation. However, it's still possible to get it if a DHCP server is running on the target VM Network and configured with the gateway information. If the user actively provides the gateway information during network creation, the network controller happily accepts it. Otherwise, the network controller will create a helper job on the target network that acts as a DHCP client to get the gateway information. With the gateway IP address in mind, the network controller then sends ICMP Echo Request packets from the management network to the gateway and waits for responses.
```
w13915984028, line 129: A question, please double check:

The helper job is set with additional network annotation, which in turn creates the 2nd interface in the job POD that connects to VM network; like the storage network, LH PODs get an [additional NIC & IP](https://docs.harvesterhci.io/v1.3/advanced/storagenetwork#step-3), all the related DHCP/ping are done on this interface. Not from the management network.

```
job.Spec.Template.ObjectMeta.Annotations[cniv1.NetworkAttachmentAnnot] = selectedNetworks
```

starbops: I've confirmed that the DHCP client runs on the target VM Network according to the pod annotation. On the other hand, the `pingGW` call is run directly inside the manager pod instead of the helper pod, so it's apparently issued from the management network:

- Initially https://github.com/harvester/network-controller-harvester/blob/ef2e5d13f0e3a37e833e6b23328b83832319d90e/pkg/controller/manager/nad/controller.go#L387
- Periodically https://github.com/harvester/network-controller-harvester/blob/ef2e5d13f0e3a37e833e6b23328b83832319d90e/pkg/controller/manager/nad/controller.go#L360

Besides, it would be useless to check the connectivity between the target VM Network and the management network just by pinging the gateway from a random IP address on the target VM network. Please correct me if I'm wrong. Thank you.

w13915984028: Per the discussion and investigation:

It looks the connectivity between mgmt network and VM network is mostly related to REK2 downstream cluster scenairo. For other general cases, those two networks are blocked.

The ping detection looks to be better as an option for VM network which is enabled by the user, the user knows he need to make sure the infrastructure layer connects them.

Let's add detailed background and connect with rancher integratio part if possible, thanks.

jillian-maroket, line 129:
```suggestion
To check connectivity, the Harvester network controller must know the gateway IP address, which is not always specified when the VM network is created. However, this address can still be obtained if a DHCP server that is configured with the gateway information is running on the target VM network. To obtain the information, the network controller creates a helper job, which functions as a DHCP client, on the target network. Once the gateway address is obtained, the network controller sends ICMP Echo Request packets from the management network to the gateway, and waits for responses.
```

docs/networking/harvester-network.md
```diff
@@ -106,3 +106,32 @@ To create a new untagged network, go to the **Networks > VM Networks** page and
...
+
+To wrap up, the **Route Connectivity** for VM Networks is an important indicator representing the connectivity between the VM Network and the management network where the Harvester nodes live.
```
jillian-maroket, line 131:
```suggestion
In summary, **route connectivity** represents connectivity between the VM network and the management network, which the Harvester nodes are connected to.
```

docs/networking/harvester-network.md
```diff
@@ -106,3 +106,32 @@ To create a new untagged network, go to the **Networks > VM Networks** page and
...
+
+:::note
+
+If a VM Network's route connectivity is `Dhcp failed`, `Ping failed`, or `Inactive`, it doesn't mean the network is entirely unusable. It depends on what you're going to do with the network. Suppose you only want to run some workloads that should be completely isolated from any other network, including the management network where the Harvester nodes live. In that case, the VM Network is suitable for the job. Whether or not the VM Network has Internet connectivity is not the concern of the Harvester network controller.
```
jillian-maroket, line 135:
```suggestion
The states **Dhcp failed**, **Ping failed**, and **Inactive** do not imply that a VM network is completely unusable. For example, if you only want isolate certain workloads from other networks (including the management network that the Harvester nodes are connected to), a VM network can still be used. Whether a VM network has internet connectivity is not the concern of the Harvester network controller.
```

docs/networking/harvester-network.md
```diff
@@ -106,3 +106,32 @@ To create a new untagged network, go to the **Networks > VM Networks** page and
...
+## About Route Connectivity
+
+![](/img/v1.3/networking/route-connectivity.png)
+
+There are four possible cases for the **Route Connectivity** for each VM Network:
+
+- `Active`, meaning the connectivity between the VM Network and Harvester hosts via the configured gateway is confirmed.
+- `Dhcp failed`, meaning Harvester cannot get the route information via DHCP, therefore it cannot confirm the connectivity between the VM Network and Harvester hosts. Please ensure the DHCP server is configured appropriately and is L2-reachable (or routable if a DHCP relay server is provided in the target network). Otherwise, please provide the gateway IP address directly during the VM Network creation.
+- `Ping failed`, meaning Harvester is unable to send ICMP Echo Request packets. This rarely happens.
+- `Inactive`, meaning such a VM Network is not reachable (or reachable but packet loss is greater than 20%) from Harvester hosts. Please ensure the gateway is configured appropriately and is reachable via the management network where the Harvester nodes live.
```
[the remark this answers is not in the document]
jillian-maroket: We usually reserve backticks for code. UI elements are bolded.

docs/networking/harvester-network.md
```diff
@@ -80,7 +80,7 @@ The [Harvester network-controller](https://github.com/harvester/harvester-networ
     ![](/img/v1.2/networking/create-network-manual.png)
 
     :::info important
-    Harvester uses the information to verify that all nodes can access the VM network you are creating. If that is the case, the *Network connectivity* column on the **VM Networks** screen indicates that the network is active. Otherwise, the screen indicates that an error has occurred.
+    Harvester uses the information to verify that all nodes can access the VM network you are creating. If that is the case, the *Network connectivity* column on the **VM Networks** screen indicates that the network is active. Otherwise, the screen indicates that an error has occurred. Please check [the Route Connectivity section](#about-route-connectivity) for more details.
```
jillian-maroket, line 83:
```suggestion
    Harvester uses the information to verify that all nodes can access the VM network you are creating. If that is the case, the *Network connectivity* column on the **VM Networks** screen indicates that the network is active. Otherwise, the screen indicates that an error has occurred. For more information, see [the Route Connectivity section](#about-route-connectivity).
```

jillian-maroket (commented):

rrajendran17:
For dhcpFailed, do you think we have to add more information to the user to understand what caused the failure ?
Like if a dhcp server is allocated and connected to the VM network ?
I am trying to understand who configures the dhcp server and is it done manually for every cluster and do we pre allocate what vlan networks we support ?

rrajendran17:
I believe dhcp server configuration per cluster is infra related and user will not have much control over it when creating vm networks.But I believe more understanding of the dhcp server configuration will make sure the user will configure the supported vm vlan networks to avoid dhcp connectivity failure.

I believe the underlying DHCP server configuration is done on the switch connected to the uplink NIC, and the switch should be configured as trunk port with various vlans supported to accept tagged traffic and DHCP server must be configured with ip pools (subnet per vlan) for various vlans supported.If this is not done correctly or if user configures a vlan vm network not supported by the dhcp server, then route connectivity for dhcp will fail.

starbops:
@rrajendran17 Thanks for the comments.

> My understanding is, the underlying DHCP server configuration is done on the switch connected to the uplink NIC, and the switch should be configured as trunk port with various vlans supported to accept tagged traffic and DHCP server must be configured with ip pools (subnet per vlan) for various vlans supported.If this is not done correctly or if user configures a vlan vm network not supported by the dhcp server, then route connectivity for dhcp will fail.

It's not that complicated. The user can set up a DHCP server for each VM network; the DHCP servers are not required to be accessible by the Harvester nodes from the management network. This is because the Harvester network controller creates a pod running the DHCP client to get the network information, including the gateway IP address for each VM network. The pod is attached to the bridge that connects to the VM network. However, when checking the connectivity, the ICMP echo requests indeed come from the Harvester nodes via the management network. That's the difference between the two-step operation.

It's okay for users to create a VM network without a DHCP server running. However, they need to provide the network information, i.e., the subnet and gateway IP, when creating the VM network (the so-called "manual" mode). The Harvester network controller can then check the route connectivity as usual.

IMO, the most important thing from users' perspectives is to make sure the VM network is routable from the management network. This ensures some of the functionalities, like load balancers, work as expected, and that's why the check is called "route connectivity." This also implies that if users don't care about those functionalities and want to create a VM network isolated from the management network, they can totally ignore the inactive/dhcpFailed route connectivity.

docs/networking/harvester-network.md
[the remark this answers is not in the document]
starbops: I have to elaborate on the **Dhcp failed** part. The previous version is not accurate and causes confusion. Please take a look. Thank you.

PR #612, opened 2024-07-17
removed usued prerequisites

FrankYang0529:
VM can have different hostnames with VM names after https://github.com/harvester/harvester/issues/4396.


bk201 (approved):

Vicente-Cheng (approved):
LGTM! Thanks for the update

PR #615, opened 2024-07-26
Add additional-guest-memory-overhead-ratio setting

w13915984028:
Issue:

https://github.com/harvester/harvester/issues/5768

https://github.com/harvester/harvester/issues/6293 (about the influence of memory overhead against resource quota)


docs/vm/create-vm.md
```diff
@@ -230,6 +230,43 @@ In order to meet the scenario requirements of more users, the `RunStrategy` fiel
 - Stop: There will be no VM instance. If the guest is already running, it will be stopped. This is the same behavior as `Running: false`.
 
 
+### Reserved Memory
+
+You can set a customized value for the VM, it depends on many factors
+
+- the numuber of devices like network nics, hot-plug volumes
+
+If no value is configured, Harvester will set a default value according to rollowing rules.
```
jillian-maroket, line 239:
```suggestion
When configuring a value for the VM, consider the number of devices (for example, network interfaces and hot-plug volumes) and other factors.

If you do not specify a value, Harvester assigns one based on the following rules:
```

docs/vm/create-vm.md
```diff
@@ -230,6 +230,43 @@ In order to meet the scenario requirements of more users, the `RunStrategy` fiel
...
+
+```
+  Expected VM Memory      Default Reserved Memory
+   100M ...  512M                  100M
+   512M ...  1G                    128M
+   1G   ...  2.5G                  256M
+   2.5G ...  10G             10%, round to M
+   10G  ...                         1G
+```
```
jillian-maroket, line 248:
```suggestion
| Expected VM Memory | Default Reserved Memory |
| --- | --- |
| 100 to 512 MiB | 100 MiB |
| 512 MiB to 1 GiB | 128 MiB |
| 1 to 2.5 GiB | 256 MiB |
| 2.5 to 10 GiB | 10%, rounded to the nearest MiB |
| 10 GiB | 1 GiB |
```

docs/vm/create-vm.md
```diff
@@ -230,6 +230,43 @@ In order to meet the scenario requirements of more users, the `RunStrategy` fiel
...
+
+For example, one VM is configured with 2GiB memory from the [Harvester UI](https://docs.harvesterhci.io/v1.3/vm/index#how-to-create-a-vm) and the `Reserved Memory` is not set, the default value of the `Reserved Memory` is 256MiB, the Guest OS gets 1792MiB memory.
+
+The final virtual machine object created on the cluster has following parameters:
```
jillian-maroket, line 252:
```suggestion
Example:

A VM that you created using the [Harvester UI](https://docs.harvesterhci.io/v1.3/vm/index#how-to-create-a-vm) has 2 GiB of memory but no set value for **Reserved Memory**. Harvester uses the default **Reserved Memory** value of 256 MiB and allocates 1792 MiB of memory to the guest OS.

The final VM object created on the cluster has the following parameters:
```

jillian-maroket (commented):
Initial review done

w13915984028:
@jillian-maroket The original PR was titled with `WIP` as it was still in progress. Now the PR is ready for review, please take a new look, thanks.

w13915984028:
The comment from code review by @bk201  is worthy to check, it is better to have some benchmarks to support our suggestions on the document.

https://github.com/harvester/harvester/pull/6438#pullrequestreview-2296005020

docs/advanced/settings.md
```diff
@@ -299,6 +299,126 @@ A VM that is configured to use 2 CPUs (equivalent to 2,000 milliCPU) can consume
 }
 ```
 
+### `additional-guest-memory-overhead-ratio`
+
+**Definition**: The ratio to futher tune the VM `memory overhead`.
+
+Each VM is configured with a memory value, this memory is targeted for the VM guest OS to see and use. In Harvester, the VM is carried by a Kubernetes POD. The memory limitation is achieved by Kubernetes [Resource requests and limits of Pod and container](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container). Certain amount of memory is required to simulate and manage the `CPU/Memory/Storage/Network/...` for the VM to run. Harvester and KubeVirt summarize such additional memory as the VM `Memory Overhead`. The `Memory Overhead` is computed by a complex formula formula. However, sometimes the OOM(Out Of Memory) can still happen and the related VM is killed by the Harvester OS, the direct cause is that the whole POD/Container exceeds its memory limits. From practice, the `Memory Overhead` varies on different kinds of VM, different kinds of VM operating system, and also depends on the running workloads on the VM.
```
ibrokethecloud, line 306:
```suggestion
Each VM is configured with a memory value, this memory is targeted for the VM guest OS to see and use. In Harvester, the VM is run in a virt-launcher pod. CPU/VM resource limits are translated and applied to the launcher pod[Resource requests and limits of Pod and container](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container). Kubevirt ensures certain amount of memory is reserved in the pod for managing the virtualisation process. Harvester and KubeVirt summarize such additional memory as the VM `Memory Overhead`. The `Memory Overhead` is computed by a complex formula formula. However, sometimes the OOM(Out Of Memory) can still happen and the related VM is killed by the Harvester OS, the direct cause is that the whole POD/Container exceeds its memory limits. From practice, the `Memory Overhead` varies on different kinds of VM, different kinds of VM operating system, and also depends on the running workloads on the VM.
```

docs/advanced/settings.md
```diff
@@ -299,6 +299,126 @@ A VM that is configured to use 2 CPUs (equivalent to 2,000 milliCPU) can consume
...
+This setting is for more flexibly tuning the VM `Memory Overhead`.
+
+**Default values**: `"1.5"`
+
+Valid values: `""`, `"0"` and from `"1.0"` to `"10.0"`.
+
+A VM that is configured to have `1 CPU, 2 Gi Memory, 1 Volume and 1 NIC` will get around `240 Mi` `Memory Overhead` when the ratio is `"1.0"`. When the ratio is `"1.5"`, the `Memory Overhead` is `360 Mi`. When the ratio is `"3"`, the `Memory Overhead` is `720 Mi`.
+
+A VM that is configured to have `1 CPU, 64 Gi Memory, 1 Volume and 1 NIC` will get around `250 Mi` `Memory Overhead` when the ratio is `"1.0"`. The VM memory size does not have a big influence on the computing of `Memory Overhead`.
+
```
ibrokethecloud, line 317: can we please also explain why with a 64G VM the mem reserved is still 250M when users may wonder it would be higher considering a 2G VM got 240M

bk201:
@w13915984028 can you check gaurav's comment? The PR has been pending for a while and we should merge this.

w13915984028:
@bk201 I will rebase it to add v1.5 branch and address the comments today.

For other reviewers, please also spend a while to comment or approve, thanks.

docs/advanced/settings.md
w13915984028: When 2 VMs have different RAM size and same other params, the final `overhead` has tiny differences:

https://github.com/kubevirt/kubevirt/blob/1466b658f78b9b8bb9517ffb6dafd4b777f33fe6/pkg/virt-controller/services/renderresources.go#L313

```
	// Add the memory needed for pagetables (one bit for every 512b of RAM size)
	pagetableMemory := resource.NewScaledQuantity(vmiMemoryReq.ScaledValue(resource.Kilo), resource.Kilo)
	pagetableMemory.Set(pagetableMemory.Value() / 512)
	overhead.Add(*pagetableMemory)
```

w13915984028: Updated the document to add 
>The overhead of `guest OS pagetables` needs one bit for every 512b of RAM size.

docs/advanced/settings.md
w13915984028: Updated, thanks.

bk201 (approved):

PR #616, opened 2024-07-30
Expand description of overcommit-config setting

tserong:
@bk201 asked for review of the definition of overcommit-config. I think it can be an irritatingly difficult thing to describe, so I've tried to expand what's there to (hopefully!) make it a bit clearer. If this looks OK to everyone, I'll update the versioned docs to match too.

### DIFF
--- a/docs/advanced/settings.md
+++ b/docs/advanced/settings.md
@@ -275,13 +275,19 @@ Changes to the server address list are applied to all nodes.
 
 ### `overcommit-config`
 
-**Definition**: Percentage of the maximum amount of CPU, memory, and storage resources that a specific VM can use.
+**Definition**: Percentage of physical CPU, memory, and storage allocatable for VM usage.
 
-This setting allows you to schedule additional VMs even when the Harvester cluster's physical resources are already fully utilized.
+Setting these values greater than 100% will allow multiple virtual machines to be scheduled even when physical resources would notionally be fully allocated. For example, with the default CPU overcommit of 1600%, memory overcommit of 150%, and storage overcommit of 200%, it will be possible to schedule:
+
+- 16x the number of physical CPUs on a host for VM usage.
+- 1.5x the amount of physical RAM on a host for VM usage.
+- 2x the amount of physical storage in Longhorn for VM storage.
+
+Overcommit is commonly used to optimize physical resource allocation when most VMs are not often expected to hit their allocated limits for CPU, memory and storage.
 
 **Default values**: `{ "cpu":1600, "memory":150, "storage":200 }`
 
-If the VM can use a maximum of 1,600 millicpu (`1600m`) and the overcommit value is 1600% (`"cpu":1600`), Harvester only requests 100 millicpu from the Kubernetes scheduler.
+If a VM is configured to use 2 CPUs (i.e. 2,000 milicpus of `2000m`), and the overcommit value is 1600%, Harvester only requests 2000/16 = 125 milicpu from the Kubernetes scheduler.
 
 **Example**:
 
--- a/docs/advanced/settings.md
+++ b/docs/advanced/settings.md
@@ -275,19 +275,19 @@ Changes to the server address list are applied to all nodes.
 
 ### `overcommit-config`
 
-**Definition**: Percentage of physical CPU, memory, and storage allocatable for VM usage.
+**Definition**: Percentage of physical compute, memory, and storage resources that can be allocated for VM use.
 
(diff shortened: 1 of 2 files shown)


tserong:
Preview is at https://66a8a64fc3a7bc142ae5f954--harvester-preview.netlify.app/v1.4/advanced/index#overcommit-config

docs/advanced/settings.md
```diff
@@ -275,13 +275,19 @@ Changes to the server address list are applied to all nodes.
...
+- 16x the number of physical CPUs on a host for VM usage.
+- 1.5x the amount of physical RAM on a host for VM usage.
+- 2x the amount of physical storage in Longhorn for VM storage.
+
+Overcommit is commonly used to optimize physical resource allocation when most VMs are not often expected to hit their allocated limits for CPU, memory and storage.
 
 **Default values**: `{ "cpu":1600, "memory":150, "storage":200 }`
 
-If the VM can use a maximum of 1,600 millicpu (`1600m`) and the overcommit value is 1600% (`"cpu":1600`), Harvester only requests 100 millicpu from the Kubernetes scheduler.
+If a VM is configured to use 2 CPUs (i.e. 2,000 milicpus of `2000m`), and the overcommit value is 1600%, Harvester only requests 2000/16 = 125 milicpu from the Kubernetes scheduler.
```
tserong, line 290: This should probably also point out that a VM configured to use 2 CPUs will still be able to use a whole 2 CPUs worth of processing power, provided such is available on the host, i.e. with overcommit in place, Harvester requests a small value from the k8s scheduler so that the VM will still schedule on a heavily loaded machine...  But if the system is not heavily loaded, that VM will happily be able to use its whole allocation. We just need to figure out how to say that using many less words :-)

docs/advanced/settings.md
```diff
@@ -275,13 +275,19 @@ Changes to the server address list are applied to all nodes.
...
+**Definition**: Percentage of physical CPU, memory, and storage allocatable for VM usage.
 
-This setting allows you to schedule additional VMs even when the Harvester cluster's physical resources are already fully utilized.
+Setting these values greater than 100% will allow multiple virtual machines to be scheduled even when physical resources would notionally be fully allocated. For example, with the default CPU overcommit of 1600%, memory overcommit of 150%, and storage overcommit of 200%, it will be possible to schedule:
+
+- 16x the number of physical CPUs on a host for VM usage.
+- 1.5x the amount of physical RAM on a host for VM usage.
+- 2x the amount of physical storage in Longhorn for VM storage.
+
+Overcommit is commonly used to optimize physical resource allocation when most VMs are not often expected to hit their allocated limits for CPU, memory and storage.
```
jillian-maroket, line 286:
```suggestion
Overcommitting is used to optimize physical resource allocation, particularly when VMs are not expected to fully consume the allocated resources most of the time. Setting values greater than 100% allows scheduling of multiple VMs even when physical resources are notionally fully allocated. 
```

jillian-maroket: I suggest moving the sample outcomes to the next section.

docs/advanced/settings.md
```diff
@@ -275,13 +275,19 @@ Changes to the server address list are applied to all nodes.
...
 
 **Default values**: `{ "cpu":1600, "memory":150, "storage":200 }`
```
jillian-maroket, line 288:
```suggestion
**Default values**: `{ "cpu":1600, "memory":150, "storage":200 }`

With the default values, it would be possible to schedule the following:

- 16x the number of physical CPUs on a host
- 1.5x the amount of physical RAM on a host
- 2x the amount of physical storage in Longhorn
```

docs/advanced/settings.md
jillian-maroket, line 290:
```suggestion
A VM that is configured to use 2 CPUs (equivalent to 2,000 milliCPU) can consume the full allocation as long as the resources are available. However, if the host is running heavy workloads and an overcommit value is set (for example, 1600%), Harvester only requests 125 milliCPU from the Kubernetes scheduler (2000/16 = 125 milliCPU).
```

jillian-maroket (commented):
Thanks for the helpful details. Let me know if you have concerns about the changes.

tserong:
That's excellent, thank you @jillian-maroket!

jillian-maroket (approved):
Thanks, Tim!

bk201 (approved):
LGTM, thanks!

ibrokethecloud (approved):

PR #626, opened 2024-08-20
doc(setting): add upgrade-config setting

starbops:
Add description for the newly introduced `upgrade-config` setting and its sub-options.

Related issues:

harvester/harvester#3059
harvester/harvester#6028

### DIFF
--- a/docs/advanced/settings.md
+++ b/docs/advanced/settings.md
@@ -503,6 +503,62 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
 https://your.upgrade.checker-url/v99/checkupgrade
 ```
 
+### `upgrade-config`
+
+**Definition**: Upgrade-related configuration.
+
+**Default value**: `{"imagePreloadOption":{"strategy":{"type":"sequential"}}}`
+
+**Supported options and fields**:
+
+- `imagePreloadOption`: Options for the image preloading phase.
+  
+  The full ISO contains the core operating system components and all required container images. Harvester can preload these container images to each node during installation and upgrades. When workloads are scheduled to management and worker nodes, the container images are ready to use.
+
+- `strategy`: Image preload strategy.
+
+- `type`: Type of image preload strategy.
+
+    - `sequential`: Harvester preloads the container images from the target ISO to each node. This is the default option.
+
+    - `skip`: Harvester does not preload the container images from the target ISO to each node. **Do not use this option in production environments.**
+
+      :::info important
+
+      If you decide to use `skip`, ensure that the following requirements are met:
+
+      - You have a private container registry that contains all required images.
+      - Your cluster has high-speed internet access and is able to pull all images from Docker Hub when necessary. 
+        
+      Note any potential internet service interruptions and how close you are to reaching your [Docker Hub rate limit](https://www.docker.com/increase-rate-limits/). Failure to download any of the required images may cause the upgrade to fail and may leave the cluster in a middle state.
+    
+      :::
+
+    - `parallel` (**experimental**): Nodes preload images in batches. You can adjust this using the `concurrency` option.
+
+- `concurrency`: Number of nodes that can simultaneously preload images. This option takes effect only when `type` is set to `parallel`.
+
+  The default value is `0`, which is equivalent to following the cluster's node counts. Using `0` allows the system to dynamically follow the scale of the cluster. Values higher than the cluster's node counts are treated as `0`, while lower values are considered invalid and are rejected by Harvester.
+
+  :::note
+
+  Harvester deploys an upgrade-repo service on the cluster that serves as an HTTP server for nodes that need to preload the container images. When a `concurrency` value is set, each batch of nodes downloads the container images from this upgrade-repo in parallel. Because of this, you must consider the speed of the Harvester management network and the read speed of the default disk for Longhorn.
+
+  :::
+
+**Example**:
+
+```json
+{
+  "imagePreloadOption": {
+    "strategy": {
+      "type": "parallel",
+      "concurrency": 2
+    }
+  }
+}
+```
+
 ### `vip-pools`
 
 **Versions**: Deprecated as of v1.2.0 (Use [IP pools](../networking/ippool.md) instead.)
--- a/docs/upgrade/automatic.md
+++ b/docs/upgrade/automatic.md
@@ -39,6 +39,10 @@ For the Harvester & Rancher support matrix, please visit our website [here](http
 
 :::
 
+## Before starting an upgrade
+
+Check out the available [`upgrade-config` setting](../advanced/settings.md#upgrade-config) to tweak the upgrade strategies and behaviors that best suit your cluster environment.
+
 ## Start an upgrade
 
 :::caution


bk201 (approved):

docs/advanced/settings.md
```diff
@@ -503,6 +503,61 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+
+**Definition**: Image preload strategy.
+
+###### `type`
+
+**Definition**: Type of the image preload strategy.
+
+There are three available types:
+
+- `skip`: Nodes will not preload images
```
w13915984028, line 539: Nodes will not preload images
->
Harvester will not preload the contaimer images from the target ISO to each node.

docs/advanced/settings.md
```diff
@@ -503,6 +503,61 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+- `sequential`: Nodes will preload images one by one. This is the default.
```
w13915984028, line 540: Harvester will preload the contaimer images from the target ISO to each node. This is the default option.

docs/advanced/settings.md
```diff
@@ -503,6 +503,61 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+- `parallel`: Nodes will preload images in batches. This can be fine-tuned by the `concurrency` option.
+
+:::important
+
+Please do not use `skip` if your cluster is air-gapped. Harvester upgrade will fail if no new images are preloaded to the nodes or downloadable from any registry.
```
w13915984028, line 545: Please recheck this statement.

When you choose `skip` please make sure:
(1) You have set a `container-registry` and all the required images are added to this registry in advance.

This looks to be the best match scenario for this option. And, Harvester needs to publish a full image list of each Release e.g. to the suggested `version.yaml`, to help user prepare the image.

A furhter enhancement will be: If user sets this option and clicks `Upgrade`, Harvester checks the container registry and the `version.yaml`,  make sure all images are ready.

(2) You plan to pull all images from docker hub when images are required. Be sure your cluster has a high speed internet access. And also notice the potential internet interruption or docker [increase-rate-limits](https://www.docker.com/increase-rate-limits/) .

The failure of downloading even one image may cause the failure of the whole upgrade and may leave the cluster in a middle state.

This is not recommended on production clusters.

docs/advanced/settings.md
```diff
@@ -503,6 +503,61 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+###### `concurrency`
+
+**Definition**: Batch size of image preloading in terms of node count.
+
+Customize the number of nodes that can start preloading images simultaneously. The default value is `0`, which is equivalent to setting it with the cluster's node counts. One benefit of setting it to `0` is it dynamically follows the scale of the cluster. Values higher than the cluster's node counts are treated as `0`; values lower than 0 are invalid and will be rejected by Harvester.
+
+:::note
+
+The `concurrency` option only takes effect when the image-preload strategy `type` is set to `parallel`.
+
```
w13915984028, line 558: Add another:

Harvester deploys a `cluster-repo` service on the cluster to serve as a http server for nodes to preload the image. When `concurrency` is set,  certain number of nodes will download the Harvester ISO from this `cluster-repo` in parallel,  following factors need to be considered:
(1) The Harvester management network speed.
(2) The writing speed of the default disk for Longhorn.

When your cluster has 20+ nodes, suggest to set the `concurrency` as `5`.


note: Before we have tested this on a at least medium sized cluster, it is better to suggest a conservative number, rather than 0.

docs/advanced/settings.md
```diff
@@ -503,6 +503,61 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+      "concurrency": 2
+    }
+  }
+}
+```
+
+#### `imagePreloadOption`
+
+**Definition**: Options for the image-preloading phase.
+
```
w13915984028, line 528: Add two more lines:

Like the new installation, Harvester tends to load all the pre-packed container images in the Harvester ISO to each node. When workloads are scheduled to any node (except the witness node), the container images are ready to use. 

This option provides more flexibilities for different scenairos.

w13915984028 (commented):
For customer convenience, I have following suggestions, thanks.

jillian-maroket:
@starbops I'll start my review after you check Jian's comments.

Also, I noticed that you used separate headings for the options. We now use an unordered list for such information (within the section dedicated to the setting). Is there any reason you decided to structure the content this way? Also, are all of the options required?

docs/advanced/settings.md
starbops: We already seem to have the image list published in previous releases. Adding those image names to `version.yaml` during packaging should be fairly easy. Thanks for the suggestion.

docs/advanced/settings.md
starbops: If we don't have data from experiments about the appropriate concurrency value, IMO we should not publish any concrete number in the documentation. Maybe just add a "this is an experimental function" warning?

starbops:
Hi @w13915984028 thank you for the suggestion. I've updated the content, PTAL.

Hi @jillian-maroket because each option under the `upgrade-config` setting deserves many content to describe its purpose, usage, and potential risks, I put them into dedicated sections. Besides, I'm unsure whether I can put that many kinds of markdown formats under an unordered list. If you have idea about this, I'm happy to change the current structure. Thank you.

w13915984028 (approved):
LGTM, thanks.

jillian-maroket:
@starbops Here is the recommended structure:

### `upgrade-config`

**Definition**: Upgrade-related configuration.

**Default value**: `{"imagePreloadOption":{"strategy":{"type":"sequential"}}}`

**Supported options and fields**:

- `imagePreloadOption`: Options for the image preloading phase.

    The full ISO contains the core operating system components and all required container images. Harvester can preload these container images to each node during installation and upgrades. When workloads are scheduled to management and worker nodes, the container images are ready to use.

- `strategy`: Image preload strategy.

- `type`: Type of image preload strategy.

    - `sequential`: Harvester preloads the container images from the target ISO to each node. This is the default option.

    - `skip`: Harvester does not preload the container images from the target ISO to each node. **Do not use this option in production environments.**

        :::info important

        If you decide to use `skip`, ensure that the following requirements are met:

        - You have a private container registry that contains all required images.
        - Your cluster has high-speed internet access and is able to pull all images from Docker Hub when necessary. 
        
        Note any potential internet service interruptions and how close you are to reaching your [Docker Hub rate limit](https://www.docker.com/increase-rate-limits/). Failure to download any of the required images may cause the upgrade to fail and may leave the cluster in a middle state.
    
        :::

    - `parallel` (**experimental**): Nodes preload images in batches. You can adjust this using the `concurrency` option.

- `concurrency`: Number of nodes that can simultaneously preload images. This option takes effect only when `type` is set to `parallel`.

    The default value is `0`, which is equivalent to following the cluster's node counts. Using `0` allows the system to dynamically follow the scale of the cluster. Values higher than the cluster's node counts are treated as `0`, while lower values are considered invalid and are rejected by Harvester.

    :::note

    Harvester deploys a cluster-repo service on the cluster that serves as an HTTP server for nodes that need to preload the image. When a `concurrency` value is set, each batch of nodes downloads the Harvester ISO from this cluster-repo in parallel. Because of this, you must consider the speed of the Harvester management network and the write speed of the default disk for Longhorn.

    :::

**Example**:

```json
{
  "imagePreloadOption": {
    "strategy": {
      "type": "parallel",
      "concurrency": 2
    }
  }
}
```

starbops:
Thank you @jillian-maroket. I slightly changed the wording in the last "note" section:

- cluster-repo -> upgrade-repo
  A `harvester-cluster-repo` pod and service is already out there, we should better use another name for the upgrade repo VM.
- Harvester ISO -> container images
  ISO image downloading happens before the image-preload phase. The nodes download the container images here.
- write -> read
  The read operations occur more often than writes in the upgrade repo VM because nodes download container images from it.

Please take a look.

docs/advanced/settings.md
```diff
@@ -503,6 +503,62 @@ This setting can only be used if the `upgrade-checker-enabled` setting is set to
...
+    - `parallel` (**experimental**): Nodes preload images in batches. You can adjust this using the `concurrency` option.
+
+- `concurrency`: Number of nodes that can simultaneously preload images. This option takes effect only when `type` is set to `parallel`.
+
+    The default value is `0`, which is equivalent to following the cluster's node counts. Using `0` allows the system to dynamically follow the scale of the cluster. Values higher than the cluster's node counts are treated as `0`, while lower values are considered invalid and are rejected by Harvester.
+
+    :::note
+
+    Harvester deploys a upgrade-repo service on the cluster that serves as an HTTP server for nodes that need to preload the container images. When a `concurrency` value is set, each batch of nodes downloads the container images from this upgrade-repo in parallel. Because of this, you must consider the speed of the Harvester management network and the read speed of the default disk for Longhorn.
```
jillian-maroket, line 545:
```suggestion
    Harvester deploys an upgrade-repo service on the cluster that serves as an HTTP server for nodes that need to preload the container images. When a `concurrency` value is set, each batch of nodes downloads the container images from this upgrade-repo in parallel. Because of this, you must consider the speed of the Harvester management network and the read speed of the default disk for Longhorn.
```

jillian-maroket (approved):
@starbops The changes work for me so I'll approve now. Just please correct the grammatical error that I flagged ("a" > "an"). Thanks!

docs/advanced/settings.md
starbops: Updated, thank you!

PR #646, opened 2024-09-30
Revise suggestion value about support bundle

Yu-Jack:
Related Issue 
- https://github.com/harvester/harvester/issues/1646

The minimum value should be `1` for these two settings.

### DIFF
--- a/docs/advanced/settings.md
+++ b/docs/advanced/settings.md
@@ -469,9 +469,9 @@ The process is considered to have failed when the data collection and file packi
 
 **Versions**: v1.3.0 and later
 
-**Definition**: Number of minutes Harvester waits before deleting a support bundle that has been packaged but not downloaded (either deliberately or unsuccessfully) or retained. 
+**Definition**: Number of minutes Harvester waits before deleting a support bundle that has been packaged but not downloaded (either deliberately or unsuccessfully) or retained.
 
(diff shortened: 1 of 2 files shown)


bk201 (approved):

albinsun (approved):
LGTM.


Note that we found `0` is also accesptable (and convert to default value `30`) but the definition is not mentioned in doc like [support-bundle-timeout](https://docs.harvesterhci.io/v1.4/advanced/index/#support-bundle-timeout).
Should open an enhancement if it's not expected.

image

Yu-Jack:
> LGTM.
> 
> Note that we found `0` is also accesptable (and convert to default value `30`) but the definition is not mentioned in doc like [support-bundle-timeout](https://docs.harvesterhci.io/v1.4/advanced/index/#support-bundle-timeout). Should open an enhancement if it's not expected.
> 
> image

If we set it to 0, we will recognize it as default value In current logic. It's expected behavior when we're developing this at that moment.

But, from current understanding, I suggest that we need to define a set of rules about these kind of setting. So, I'll update this documentation to let it match what we have.

albinsun (approved):

jillian-maroket (commented):
I made a few minor changes to the draft.

jillian-maroket (approved):

PR #647, opened 2024-10-08
docs(backup-restore): add v1.4.0 without manual uploading same image

FrankYang0529:
https://github.com/harvester/harvester/issues/6721

### DIFF
--- a/docs/vm/backup-restore.md
+++ b/docs/vm/backup-restore.md
@@ -88,7 +88,7 @@ _Available as of v1.0.0_
 
 Users can now restore a new VM on another cluster by leveraging the VM metadata & content backup feature.
 
-:::info prerequisites
+#### Prerequisites
 
 To restore a new VM from Harvester cluster above v1.4.0,
 users don't need to manually upload the VM images to the new cluster.
--- a/docs/vm/backup-restore.md
+++ b/docs/vm/backup-restore.md
@@ -90,7 +90,19 @@ Users can now restore a new VM on another cluster by leveraging the VM metadata 
 
 :::info prerequisites
 
-You must manually configure the virtual machine images with the same name on the new cluster first, otherwise the virtual machines will be failed to recover.
+To restore a new VM from Harvester cluster above v1.4.0,
+users don't need to manually upload the VM images to the new cluster.
+The controller will automatically sync the VM images to the new cluster.
+
+There are two limitations for automatically sync the virtual machine images to the new cluster:
+
+1. There is a virtual machine image with the same name on the new cluster.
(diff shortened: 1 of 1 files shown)


bk201 (approved):
lgtm

docs/vm/backup-restore.md
```diff
@@ -90,7 +90,19 @@ Users can now restore a new VM on another cluster by leveraging the VM metadata
...
+
+1. There is a virtual machine image with the same name on the new cluster.
+2. There is a virtual machine image with the same display name on the new cluster.
+
+To restore a new VM from Harvester cluster below v1.4.0.
+You must manually configure the virtual machine images with the same name on the new cluster first,
+otherwise the virtual machines will be failed to recover.
+Please check the [Upload the same VM images to a new cluster](#upload-the-same-vm-images-to-a-new-cluster) section for more details.
 
 :::
```
jillian-maroket, line 107:
```suggestion
- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, except when a virtual machine image with the same name or display name already exists on the new cluster.

- Earlier than v1.4.0: You must upload and configure the virtual machine images on the new cluster. Ensure that the image names and configuration are identical so that the virtual machines can be restored. For more information, see [Upload the same VM images to a new cluster](#upload-the-same-vm-images-to-a-new-cluster).
```

jillian-maroket: What happens when a virtual machine image with the same name or display name already exists on the new cluster? Will users encounter an error message? Will they be asked to choose between the two images?

jillian-maroket (commented):
I converted the note to a subsection (proper format) and used an unordered list (easier scanning). The limitations part does not require its own list.

docs/vm/backup-restore.md
FrankYang0529: The system will skip images with same name or display name. There will have warning message in the harvester pod.

jillian-maroket: Got it. I asked because users might wonder about how Harvester handles the exception. If you think the warning message provides enough information, we don't need to enhance the description here.

FrankYang0529: Sorry, I thought you approve the PR, so I merged it. The error message is like following. Do you think we need another PR to show more information here? Thanks.

```
skip create vm image, because there is already an image with the same display name

skip creating vm backup, because the storage class is not found
```

FrankYang0529: The warning message will also has vm image namespace / name and vm backup namespace / name. Above message is just an example.

PR #648, opened 2024-10-08
docs(backup-restore): add VM Snapshot Space Management

FrankYang0529:
https://github.com/harvester/harvester/issues/4478


bk201 (approved):

jillian-maroket (commented):
Two things to note:
- Alex wants me to use "virtual machine" instead of "VM" moving forward.
- The intro mentions both backups and snapshots but the procedures only mention snapshots. I don't know if this is accurate.

jillian-maroket (approved):
Thanks for the clarification. LGTM

PR #651, opened 2024-10-16
fix: use proper sentence

Yu-Jack:
https://github.com/harvester/harvester/issues/1710#issuecomment-2415797450

### DIFF
--- a/docs/advanced/addons/pcidevices.md
+++ b/docs/advanced/addons/pcidevices.md
@@ -152,4 +152,4 @@ Before you remove the USB device, detach it from the virtual machine and then di
 
 - Hot-plugging and replugging of USB devices is not supported. For more information, see [KubeVirt Issue #11979](https://github.com/kubevirt/kubevirt/issues/11979).
 
-- If the device path changes when you reattach the device or reboot the node, you must detach it from the virtual machine and follow the steps above to repeat enabling process.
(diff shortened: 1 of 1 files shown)


bk201 (approved):

docs/advanced/addons/pcidevices.md
```diff
@@ -152,4 +152,4 @@ Before you remove the USB device, detach it from the virtual machine and then di
 
 - Hot-plugging and replugging of USB devices is not supported. For more information, see [KubeVirt Issue #11979](https://github.com/kubevirt/kubevirt/issues/11979).
 
-- If the device path changes when you reattach the device or reboot the node, you must create a new USBDeviceClaim to enable the device.
+- If the device path changes when you reattach the device or reboot the node, you must detach it from the virtual machine and follow the steps above to repeat enabling process.
```
jillian-maroket, line 155:
```suggestion
- If the device path changes when you reattach the device or reboot the node, you must detach the device from the virtual machine and then enable passthrough again.
```

jillian-maroket: We avoid using words such as "above" and "below" because the relevant information may be relocated in the future.

jillian-maroket (commented):
Review done

PR #657, opened 2024-10-24
doc: Adjust USB Install Info

irishgordo:
* adjusting USB Harvester ISO installtion information

Resolves: doc/adjust-usb-info


irishgordo:
@Vicente-Cheng @bk201 @ibrokethecloud 

I've kept this as a draft, but wanted to gather all your thoughts around an idea of possibly putting some-sort-of a disclaimer about the tools possibly used to flash a USB device.  Additionally adding a Rust based open-source Linux tool (that allows for multiple drives) to the list of possible tools.

As with some of the tools for instance, depending on the version, there may indeed be issues that a user might encounter.

As a brief example, balenaEtcher:
- https://www.github.com/balena-io/etcher/issues/4150

PR #668, opened 2024-11-15
feat: add upgrade page for v1.3.2 to v1.4.0

FrankYang0529:
ref: https://github.com/harvester/harvester/issues/7005

### DIFF
--- a/docs/upgrade/v1-3-2-to-v1-4-0.md
+++ b/docs/upgrade/v1-3-2-to-v1-4-0.md
@@ -1,0 +1,34 @@
+---
+sidebar_position: 2
+sidebar_label: Upgrade from v1.3.2 to v1.4.0
+title: "Upgrade from v1.3.2 to v1.4.0"
+---
+
+<head>
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.4/upgrade/v1-3-2-to-v1-4-0"/>
+</head>
+
+## General information
+
+An **Upgrade** button appears on the **Dashboard** screen whenever a new Harvester version that you can upgrade to becomes available. For more information, see [Start an upgrade](./automatic.md#start-an-upgrade).
+
+For air-gapped environments, see [Prepare an air-gapped upgrade](./automatic.md#prepare-an-air-gapped-upgrade).
+
+
+## Known issues
+
+---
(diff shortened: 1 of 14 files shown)


docs/upgrade/v1-3-2-to-v1-4-0.md
```diff
@@ -0,0 +1,34 @@
...
+
+### 1. A VM with container disk can't be migrated whick makes the upgrade stuck in pre-drain status
+
+:::tip
+
+Manually stop the VMs to continue the upgrade process.
+
+:::
+
+When upgrading from v1.3.2 to v1.4.0, the upgrade process may become stuck if a VM with a container disk cannot be migrated. There is some limitation of live migration.
```
jillian-maroket, line 30:
```suggestion
### 1. The upgrade process becomes stuck in the "Pre-draining" state.

A virtual machine with a container disk cannot be migrated because of a limitation of the Live Migration feature. This causes the upgrade process to become stuck in the "Pre-draining" state.

:::tip

Manually stop the virtual machines to continue the upgrade process.

:::
```

jillian-maroket (commented):
I reorganized the information so that it follows the problem-cause-action pattern. We provide a summary of the issue in the heading, explain the root cause, and then recommend a user action.

jillian-maroket (approved):

docs/upgrade/v1-3-2-to-v1-4-0.md
```diff
@@ -0,0 +1,34 @@
...
+An **Upgrade** button appears on the **Dashboard** screen whenever a new Harvester version that you can upgrade to becomes available. For more information, see [Start an upgrade](./automatic.md#start-an-upgrade).
+
+For air-gapped environments, see [Prepare an air-gapped upgrade](./automatic.md#prepare-an-air-gapped-upgrade).
+
+
+## Known issues
+
+---
+
+### 1. A VM with container disk can't be migrated whick makes the upgrade stuck in pre-drain status
```
bk201, line 22:
```suggestion
### 1. A VM with a container disk can't be migrated which makes the upgrade stuck in pre-drain status
```

bk201 (approved):

PR #669, opened 2024-11-15
Add/update canonical links and fix issues

jillian-maroket:
I want to update the canonical links now so that the v1.5.0 doc is created with the correct links already in place. I'll update the v1.3.0 canonical links after the release.

Change log (v1.4.0):
- Add and update canonical links
- Fix capitalization issues in proper nouns (example: Nvidia -> NVIDIA)
- Apply wording changes that were suggested in the review but not implemented by the engineer (either not at all or errors were reintroduced to the file after I approved the PR)


LucasSaintarbor (approved):
LGMT. Thanks @jillian-maroket!

PR #673, opened 2024-11-25
docs: add the known issue caution on Harvester CSI driver (v1.3)

bk201:
Backport to v1.3.

The doc should be in the v1.3 branch too since it mainly affects v1.3 clusters.


Vicente-Cheng (approved):
thanks for the backport!

innobead (approved):

PR #674, opened 2024-11-25
docs: include the privilege pods info on the upgrade page

ihcsim:
This PR updates the upgrade page to include info on privileged permissions required to successfully run the upgrade operation, per https://github.com/harvester/harvester/issues/6963#issuecomment-2477583792.

ref: https://github.com/harvester/harvester/issues/6963.


jillian-maroket (approved):
LGTM. Thanks!

PR #676, opened 2024-11-26
Mention LHv2 doesn't support trim with AIO bdev driver

tserong:
Non-NVMe disks use the SPDK AIO bdev driver, which doesn't support the unmap operation.  This means attempting to use trimming (also known as discard) with these devices will result in I/O errors and will pause your virtual machines.

This PR adds details about this limitation and gives an example for how to work around it when creating ext4 filesystems in Linux VMs, and also mentions how to disable trim for NTFS in Windows VMs. For reference, the latter command was found in this forum post:

  https://www.tenforums.com/tutorials/40028-enable-disable-trim-support-solid-state-drives-windows-10-a.html

I've added this information to the existing note about size limitations of non-NVMe disks, but if this note is getting too long I guess we could always break it out into a separate section.

As with the size limitation, this issue only affects non-NVMe disks. NVMe disks using the SPDK NVMe bdev driver support unmap and thus trim/discard works just fine on those.

### DIFF
--- a/docs/advanced/longhorn-v2.md
+++ b/docs/advanced/longhorn-v2.md
@@ -73,9 +73,11 @@ The Longhorn V2 Data Engine is only available for newly created volumes. Existin
 
   :::note
 
-  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance.
+  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance and also supports advanced operations like trimming.
   
   SSDs and other non-NVMe disks are managed using the SPDK AIO bdev driver, which requires a disk size that is an *even multiple of 4096 bytes*. Non-NVMe disks that do not meet this size requirement cannot be added.
+
+  Additionally, the SPDK AIO bdev driver does not support the unmap operation. This means attempting to use trimming (also known as discard) with these devices will result in I/O errors and will pause your virtual machines. Accordingly, if you are using non-NVMe disks, you should avoid trim/discard.  For example, when creating an ext4 filesystem in a Linux VM, use `mkfs.ext4 -E nodiscard /dev/vdb` (assuming `/dev/vdb` is your device path).  On Windows VMs, trim support for NTFS can be disabled by running `fsutil behavior set disabledeletenotify NTFS 1` from the command prompt.
 
   :::
 


docs/advanced/longhorn-v2.md
```diff
@@ -73,10 +73,12 @@ The Longhorn V2 Data Engine is only available for newly created volumes. Existin
...
+  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance and also supports advanced operations like trimming.
   
   SSDs and other non-NVMe disks are managed using the SPDK AIO bdev driver, which requires a disk size that is an *even multiple of 4096 bytes*. Non-NVMe disks that do not meet this size requirement cannot be added.
 
+  Additionally, the SPDK AIO bdev driver does not support the unmap operation. This means attempting to use trimming (also known as discard) with these devices will result in I/O errors and will pause your virtual machines. Accordingly, if you are using non-NVMe disks, you should avoid trim/discard.  For example, when creating an ext4 filesystem in a Linux VM, use `mkfs.ext4 -E nodiscard /dev/vdb` (assuming `/dev/vdb` is your device path).  On Windows VMs, trim support for NTFS can be disabled by running `fsutil behavior set disabledeletenotify NTFS 1` from the command prompt.
```
bk201, line 80: A note seems too gentle, we should raise the level to caution or warning. Or we can integrate this into the "Limitations" section?

tserong: Yeah I did wonder if this wanted to be moved to another section. Happily it's not a problem for NVMe disks (i.e. anything that shows up on the host as `/dev/nvmeXXXX`), but folks trying non-NVMe SSDs will likely hit this.

jillian-maroket: As Kiefer suggested, please move this to the Limitations section and use an unordered list.

- Within the Harvester context, volumes backed by the Longhorn V2 Data Engine must be added to virtual machines as extra disks. The boot disk of each virtual machine must still be added from an image that is backed by the Longhorn V1 Data Engine.

- Harvester is unable to live-migrate virtual machines with V2 volumes attached, which means those virtual machines will be shut down during Harvester cluster upgrades. Moreover, snapshots of V2 volumes cannot be created because snapshot and restoration functionality in Harvester relies on volume cloning.

- SSDs and other non-NVMe disks are managed using the SPDK AIO bdev driver, which does not support the unmap operation. If you are using non-NVMe disks, avoid trimming the filesystem because this results in I/O errors and paused virtual machines. For example, when creating an ext4 filesystem on a Linux virtual machine, use `mkfs.ext4 -E nodiscard /dev/vdb` (assuming `/dev/vdb` is your device path). On Windows virtual machines, you can disable trimming for NTFS by running the command `fsutil behavior set disabledeletenotify NTFS 1`.

docs/advanced/longhorn-v2.md
```diff
@@ -73,10 +73,12 @@ The Longhorn V2 Data Engine is only available for newly created volumes. Existin
 
   :::note
 
-  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance.
+  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance and also supports advanced operations like trimming.
```
jillian-maroket, line 76:
```suggestion
  Harvester sets the [Longhorn disk driver](https://longhorn.io/docs/1.7.2/v2-data-engine/features/node-disk-support/) to `auto` so that NVMe disks use the SPDK NVMe bdev driver, which provides the best performance and also supports advanced operations such as trim (also known as discard).
```

docs/advanced/longhorn-v2.md
```diff
@@ -73,10 +73,12 @@ The Longhorn V2 Data Engine is only available for newly created volumes. Existin
...
   
   SSDs and other non-NVMe disks are managed using the SPDK AIO bdev driver, which requires a disk size that is an *even multiple of 4096 bytes*. Non-NVMe disks that do not meet this size requirement cannot be added.
```
jillian-maroket, line 78:
```suggestion
  SSDs and other non-NVMe disks are managed using the SPDK AIO bdev driver, which requires a disk size that is an *even multiple of 4096 bytes*. Non-NVMe disks that do not meet this size requirement cannot be added.  Additionally, the SPDK AIO bdev driver does not support the unmap operation. If you are using non-NVMe disks, avoid trimming the filesystem because this results in I/O errors and paused virtual machines.
```

docs/advanced/longhorn-v2.md
jillian-maroket: You can change :::note to :::info important (since the text block is no longer purely informational).

jillian-maroket (commented):
Let me know if you have concerns about the suggested info placement.

jillian-maroket (approved):
LGTM. Thanks, @tserong !

PR #678, opened 2024-11-28
Fix broken links flagged by Docusaurus

jillian-maroket:
Change log:
- v1.5: Added missing "."
- v1.4: Added missing "."
- v1.2: Added missing "."; removed links to pages that exist only in later versions of the doc (the corresponding features were mentioned in high-level descriptions)

> [!NOTE]
> I fixed the v1.2 doc issues to keep the build log clean. We no longer add content to that version.


tserong (approved):
LGTM!

PR #682, opened 2024-12-02
feat: add description of available parameters for volume encryption

Yu-Jack:
https://github.com/harvester/harvester/issues/6789

### DIFF
--- a/docs/image/image-security.md
+++ b/docs/image/image-security.md
@@ -54,14 +54,16 @@ Prepare the following resources:
 
   :::info important
 
+  The example contains the default YAML code for Kubernetes secrets. Aside from this, you can use [encryption options for LUKS mode](https://wiki.archlinux.org/title/Dm-crypt/Device_encryption#Encryption_options_for_LUKS_mode), which is a cryptsetup operating mode. Harvester v1.4.1 and later versions support these options, but you must verify that these are supported by your nodes.
+
+  | Option | Possible Values |
+  | --- | --- |
+  | CRYPTO_KEY_CIPHER | aes-xts-plain, aes-xts-plain64, aes-cbc-plain, aes-cbc-plain64, aes-cbc-essiv:sha256 |
+  | CRYPTO_KEY_HASH | sha256, sha384, sha512 |
+  | CRYPTO_KEY_SIZE | 256, 384, 512 |
+  | CRYPTO_PBKDF | argon2i, argon2id, pbkdf2 |
+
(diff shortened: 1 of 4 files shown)


bk201 (approved):

docs/image/image-security.md
```diff
@@ -56,6 +56,12 @@ Prepare the following resources:
 
   You can create a secret in the system namespace using kubectl or the Harvester UI (**Edit as YAML** feature). Resources in the system namespace are not displayed on the Harvester UI **Secrets** screen.
 
+  The example of secret is default yaml. From Harvester v1.4.1, we also support to accept these parameters from [cryptsetup](https://wiki.archlinux.org/title/Dm-crypt/Device_encryption#Encryption_options_for_LUKS_mode), but it eventually depends on whether your node supports them or not. 
+
+  - CRYPTO_KEY_CIPHER: aes-xts-plain, aes-xts-plain64, aes-cbc-plain, aes-cbc-plain64, aes-cbc-essiv:sha256  
+  - PBKDF: argon2i, argon2id, pbkdf2  
+  - Hash: sha256, sha384, sha512  
+  - Key Size: 256, 384, 512
```
jillian-maroket, line 64:
```suggestion
The example contains the default YAML code for Kubernetes secrets. Aside from this, you can use [encryption options for LUKS mode](https://wiki.archlinux.org/title/Dm-crypt/Device_encryption#Encryption_options_for_LUKS_mode), which is a cryptsetup operating mode. Harvester v1.4.1 and later versions support these options, but you must verify that these are supported by your nodes.

| Option | Accepted Values |
| --- | --- |
| CRYPTO_KEY_CIPHER | aes-xts-plain, aes-xts-plain64, aes-cbc-plain, aes-cbc-plain64, aes-cbc-essiv:sha256 |
| PBKDF | argon2i, argon2id, pbkdf2 |
| Hash | sha256, sha384, sha512 |
| Key Size | 256, 384, 512 |

You can create a secret in the system namespace using kubectl or the Harvester UI (**Edit as YAML** feature). Resources in the system namespace are not displayed on the Harvester UI **Secrets** screen.
```

jillian-maroket (commented):
I switched the paragraphs so that the new information is closer to the secret example. I also converted the unordered list to a table because I need labels for the information types.

Please check if items in the "Option" column are written exactly how you expect them to be used. For example, I don't know if users should use "Hash" or "CRYPTO_KEY_HASH" or "--hash".

Yu-Jack:
Hi @jillian-maroket, I just adjusted the tab because the suggested change broke the tab, thanks!

jillian-maroket (approved):
LGTM. Thanks!

PR #688, opened 2024-12-10
feat: add crawler configuration

Yu-Jack:
Related Issue: https://github.com/harvester/harvester/issues/6623

Add a new Github Action for DocSearch crawler. 

[In my testing Github Action](https://github.com/Yu-Jack/docs/actions/runs/12250639089/job/34173924809), it ran for 5 minutes. There are few errors. They're about service limitations. We can improve our section later for better searching. 

Basically, we can have 1m records based on [pricing page](https://www.algolia.com/pricing). 

TODO:

- [x] Add two new secrets

image

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,9 @@ jobs:
       contents: write
       pages: write
     steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


jillian-maroket (approved):
@Yu-Jack LGTM, thank you. Including paragraphs and list entries in the search scope will significantly improve UX. Also, we can archive some of the older doc versions as we approach the 1m record limit.

PR #689, opened 2024-12-11
Add workaround for maintenance mode

bk201:
Related issue: https://github.com/harvester/harvester/issues/7128

PR #695, opened 2024-12-24
Rewrite and relocate node maintenance warning

jillian-maroket:
I rewrote the warning added in [PR 689](https://github.com/harvester/docs/pull/689/) to improve readability. I also placed the warning below the intro paragraph so that users new to Maintenance Mode can become familiar with it first.


bk201 (approved):
Thanks for the refactor!

PR #703, opened 2025-01-07
doc(rancher): suggest managed dhcp for creating guest cluster

starbops:
This is to give users a hint that there is another way to set up DHCP for creating guest clusters.

Related issue: harvester/harvester#6469

### DIFF
--- a/docs/rancher/node/k3s-cluster.md
+++ b/docs/rancher/node/k3s-cluster.md
@@ -15,6 +15,7 @@ You can now provision K3s Kubernetes clusters on top of the Harvester cluster in
 :::note
 
 - Harvester K3s node driver is in **Tech Preview**.
+- Provisioning K3s Kubernetes clusters involves configuring the IP address of the underlying virtual machines. You can do this using a DHCP server on the VLAN network that the virtual machines are attached to. If such a server does not exist on the network, you can use the [Managed DHCP](../../advanced/addons/managed-dhcp.md) feature to configure the IP address.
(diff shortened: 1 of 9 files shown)


bk201 (approved):

docs/rancher/node/k3s-cluster.md
```diff
@@ -15,6 +15,7 @@ You can now provision K3s Kubernetes clusters on top of the Harvester cluster in
 :::note
 
 - Harvester K3s node driver is in **Tech Preview**.
+- Provisioning K3s Kubernetes clusters requires a means to configure the IP address of the underlying virtual machines. A typical way is to use DHCP. If you don't have an existing DHCP server on the VLAN network where the virtual machines are attached to, or you don't bother to spin up a new DHCP server, you can use the [Managed DHCP](../../advanced/addons/managed-dhcp.md) feature to achieve the same goal.
```
jillian-maroket, line 18:
```suggestion
- Provisioning K3s Kubernetes clusters involves configuring the IP address of the underlying virtual machines. You can do this using a DHCP server on the VLAN network that the virtual machines are attached to. If such a server does not exist on the network, you can use the [Managed DHCP](../../advanced/addons/managed-dhcp.md) feature to configure the IP address.
```

jillian-maroket (commented):
Review done

docs/rancher/node/k3s-cluster.md
starbops: I've updated the wording. PTAL. Thank you!

jillian-maroket (approved):

innobead (approved):

PR #704, opened 2025-01-13
fix broken link in lvm-local-storage.md

SlavikCA:


ihcsim:
@SlavikCA Thanks for the PR. I don't think this issue is caused by the target path. While I can't fully reproduce it, I notice that _sometimes_ the generated href is wrong, depending on how many times I hit 'refresh' or 'back' (on FF and Chrome).

E.g., `https://docs.harvesterhci.io/v1.4/host/#multi-disk-management` here:

image

versus `https://docs.harvesterhci.io/v1.4/advanced/host/#multi-disk-management`, which is wrong:

image

Comparing this with other pages, I suspect the correct href to use here is `[Multi-disk Management](../host/host.md#multi-disk-management` (with the `host.md` added).

PR #705, opened 2025-01-13
doc: update iso installation page with new cluster cidr and dhcp settings

ihcsim:
This PR updates the ISO installation page to cover new changes added to the cluster CIDRs and DHCP configuration, per the following issues:

https://github.com/harvester/harvester/issues/4254
https://github.com/harvester/harvester/issues/6960

It also updates the Harvester Configuration page to include the new cluster CIDR fields.


docs/install/iso-install.md
```diff
@@ -87,17 +87,31 @@ The following [video](https://youtu.be/X0VIGZ_lExQ) shows a quick overview of an
...
+
 1. (Optional) Configure the `DNS Servers`. Use commas as a delimiter to add more DNS servers. Leave it blank to use the default DNS server.
 
 	![config-dns-server.png](/img/v1.2/install/config-dns-server.png)
 
 1. Configure the virtual IP (VIP) by selecting a `VIP Mode`. This VIP is used to access the cluster or for other nodes to join the cluster.
 
 	:::note
-	If using DHCP to configure the IP address, you need to configure a static MAC-to-IP address mapping on your DHCP server to have a persistent virtual IP (VIP), and the VIP must be unique.
+	For DHCP setup with static MAC-to-IP address mappings configured, enter the MAC address in the provided field to fetch the unique persistent virtual IP (VIP). Otherwise, leave it blank.
```
w13915984028, line 111: A question about this statement:

In DHCP mode, even if server side has no pre-configured IP-MAC binding, can the user still enter a preferred MAC? Namely, user is free to enter any valid MAC.

w13915984028 (approved):
LGTM, thanks.

docs/install/harvester-configuration.md
```diff
@@ -73,6 +73,9 @@ install:
   vip: 10.10.0.19
   vip_hw_addr: 52:54:00:ec:0e:0b
   vip_mode: dhcp
+  cluster_pod_cidr: 172.16.0.0/16
```
bk201, line 76: I'd suggest we use default values here

w13915984028: This also reminds me below topic:

The CIDR looks to have three reserved IPv4 ranges for private network: `10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16`; other CIDR may encounter risks. @ihcsim Is this checked on Harvester-installer? thanks.

https://en.wikipedia.org/wiki/Reserved_IP_addresses

docs/install/iso-install.md
ihcsim: Good point. If user provides a non-configured MAC, DHCP will just issue a new IP. The `vip-*@mgmt-br` links end up with that particular IP-MAC pairing.

Taking my testing a bit further, if the installer sends the MAC address of another cluster's VIP, DHCP will blindly return the same VIP of that particular cluster to the installer. If the user is unaware that IP is already assigned, it can ruin the installation. At the moment, I don't know if there is an option we can send to DHCP to for example, error out if the IP is already in-use. I believe the same problem also exists in the current automatic installation modes.

w13915984028: The latter is a networking plan/management problem, not a Harvester issue. Luckily, most business customers have processes to make sure the IP/MAC... are well managed.

docs/install/harvester-configuration.md
ihcsim: @w13915984028 No, we don't have any such checks on the installer. We only make sure they are well-formed CIDRs. It will be nice if these checks can be delegated to RKE2, then RKE2 can tell us exactly what is/isn't acceptable. But AFAICT, [RKE2 also doesn't do this](https://github.com/k3s-io/k3s/blob/646e3135bc9f3afd1f18a5eb13d0a9ffb7f965a1/pkg/cli/server/server.go#L318-L329).

It might not be a bad idea to do this check in the installer, but also feel might not be necessary at this point. WDYT?

w13915984028: We may consider to add a suggestion in this document PR, recommend users to set subnets from the above 3 private networks.

docs/install/iso-install.md
ihcsim: Fixed in 99a2455f32ad0b21de0e09788fa12529a3551fc1.

docs/install/harvester-configuration.md
```diff
@@ -728,6 +731,49 @@ install:
...
+```
+
+### `install.cluster_service_cidr`
+
+**Definition**: CIDR of the Harvester services.
+
+Use this field to override the default service CIDR of 10.53.0.0/16.
+
+If this CIDR is changed, the cluster DNS IP must be within this range. See the `install.cluster_dns` field.
```
jillian-maroket, line 753:
```suggestion
:::info important

If you change this CIDR, you must ensure that the cluster DNS IP (`install.cluster_dns`) is within this range.

:::
```

docs/install/harvester-configuration.md
```diff
@@ -728,6 +731,49 @@ install:
...
+  cluster_service_cidr: 172.22.0.0/16
+```
+
+### `install.cluster_dns`
+
+**Definition**: IP of the Harvester DNS service.
+
+Use this field to override the default DNS service IP of 10.53.0.10.
+
+This IP must be within the range defined by the `cluster_service_cidr` field.
```
jillian-maroket, line 768:
```suggestion
:::info important

This IP must be within the range defined by the `cluster_service_cidr` field.

:::
```

docs/install/iso-install.md
```diff
@@ -87,17 +87,35 @@ The following [video](https://youtu.be/X0VIGZ_lExQ) shows a quick overview of an
...
+  * Service CIDR: 172.22.0.0/16
+  * Cluster DNS IP: 172.22.0.10
+
+	:::note
+  The pod CIDR and service CIDR must not overlap.
+
+  The CIDRs must be within the private IP address range of either 10.0.0.0/8, 172.16.0.0/12 or 192.168.0.0/16.
+
+  The DNS service IP must be within the service CIDR.
+	:::
```
jillian-maroket, line 106:
```suggestion
1. (Optional) Configure the CIDRs for the cluster pods and services.

    If you want to use the default values, leave the fields blank.

    ![config-cluster-cidrs.png](/img/v1.5/install/config-cluster-cidrs.png)

    :::info important
  
    The CIDR values must not overlap and must be within the private IP address range of 10.0.0.0/8, 172.16.0.0/12, or 192.168.0.0/16.

    The DNS service IP must be within the range defined by the **Service CIDR** field.
    
    :::
    
    Example of a valid CIDR configuration:

    - **Pod CIDR**: 172.16.0.0/16
    - **Service CIDR**: 172.22.0.0/16
    - **Cluster DNS IP**: 172.22.0.10
```

jillian-maroket (commented):
Just a few nits

ihcsim:
@jillian-maroket thanks - fixed in 4339e8e7bc1a61337ca526082b8476ab631131ff.

jillian-maroket (approved):

bk201 (approved):

PR #713, opened 2025-02-07
Add doc for how to install rancher-alerting-drivers

brandboat:
related to https://github.com/harvester/harvester/issues/7369, add documentation for installing rancher-alerting-drivers in Harvester v1.3, v1.4, and the latest versions.

### DIFF
--- a/docs/monitoring/harvester-monitoring.md
+++ b/docs/monitoring/harvester-monitoring.md
@@ -181,6 +181,52 @@ Fill in the required parameters and click `Create`.
 
 ![](/img/v1.2/monitoring/webhook-receiver-2.png)
 
+To set up Microsoft Teams or SMS webhooks, first install the rancher-alerting-drivers app using the following commands:
+
+```shell
+helm repo add rancher-charts https://charts.rancher.io/
+helm repo update
+helm install rancher-charts/rancher-alerting-drivers \
+  --set sachet.enabled=false \  # Set to true if you want to use SMS Webhook
+  --set prom2teams.enabled=true \ # Set to true if you want to use MS Teams Webhook
+  --namespace cattle-monitoring-system \
+  --generate-name
+```
+
+For detailed configuration instructions, see [Rancher Monitoring Receiver Configuration](https://ranchermanager.docs.rancher.com/reference-guides/monitoring-v2-configuration/receivers) in the Rancher documentation.
+
+If your environment does not have direct internet access (air-gapped), you must manually download the Helm chart and related container images, and then upload them to the Harvester cluster.
+
+1. Download the rancher-alerting-drivers Helm chart and package it.
+
+    ```
+    helm pull rancher-charts/rancher-alerting-drivers --version <VERSION>
+    ```
+
+1. Download the required images.
+
+    ```
+    docker save -o sachet.tar rancher/mirrored-messagebird-sachet:<VERSION>
+    docker save -o prom2teams.tar rancher/mirrored-idealista-prom2teams:<VERSION>
+    ```
+
+1. Upload the chart and images to the Harvester cluster.
+
+1. Load the images on all Harvester nodes.
+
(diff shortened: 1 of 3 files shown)


docs/monitoring/harvester-monitoring.md
```diff
@@ -181,6 +181,17 @@ Fill in the required parameters and click `Create`.
 
 ![](/img/v1.2/monitoring/webhook-receiver-2.png)
 
+To set up MS Teams or SMS Webhook, first install rancher-alerting-drivers using the following commands. For detailed configuration instructions, please refer to the [Rancher Monitoring Receiver Configuration](https://ranchermanager.docs.rancher.com/reference-guides/monitoring-v2-configuration/receivers)
+```shell
```
w13915984028, line 185: please also list to output objects of this command, user can check the result;

meanwhile, in the continuous upgrade path, how to make sure they are upgraded? and for air-gapped env, how to? could we also list the related container images.

brandboat: > continuous upgrade path – how do we ensure they are upgraded?

Since the doc use helm, users need to upgrade rancher-alerting-drivers manually.

> air-gapped environments – how should we handle this?

The only way I know is packaging rancher-alerting-drivers into the ISO.

Actually, I’m wondering—if rancher-alerting-drivers is part of rancher-monitoring, should we consider adding it to managedchart, I mean add it to harvester addon? That way, we could handle the upgrades ourselves more easily. But I'm not sure if this is the right way, thoughts? cc. @ibrokethecloud, @bk201

w13915984028: As a part of rancher-monitoring stack is the preferred solution.

If Rancher does not pack it, we can do it via https://github.com/harvester/harvester-installer/blob/master/scripts/hack/patch-rancher-monitoring , and add related container images to ISO.

In the ugprade path, we take care it too.

brandboat: > If Rancher does not pack it, we can do it via https://github.com/harvester/harvester-installer/blob/master/scripts/hack/patch-rancher-monitoring , and add related container images to ISO.
>
> In the ugprade path, we take care it too.

Hi @w13915984028. That sounds like a great idea! But I believe this requires further discussion and could be considered for the future. For now, let’s enhance the documentation. I will include content on how to add the related image to the ISO. Regarding the upgrade part, unfortunately, users will have to manage that on their own for the time being.

w13915984028: NP. Please add a :::note::: to describe those limitations. thanks.

brandboat: Updated, please take a look when you have a chance, thanks!

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

docs/monitoring/harvester-monitoring.md
```diff
@@ -181,6 +181,52 @@ Fill in the required parameters and click `Create`.
...
+    docker load -i prom2teams.tar
+    ```
+
+1. Install rancher-alerting-drivers within the Harvester cluster.
+
+:::note
+
+`rancher-alerting-drivers` is not part of the Harvester project, meaning Harvester does not manage its upgrades. Users have to maintain and upgrade it manually.
+
+:::
```
jillian-maroket, line 228:
```suggestion
:::info important

Harvester does not manage upgrades of the rancher-alerting-drivers app, which is not part of the Harvester project. You must upgrade the app manually.

:::
```

jillian-maroket (commented):
Just a few nits

brandboat:
Addressed comments and rebased from main branch, @jillian-maroket please take a look when you are available, thank you!

jillian-maroket (approved):
LGTM. Thanks!

PR #715, opened 2025-02-11
fix: add v1.4.1 known issue: upgrade stuck in "Waiting Reboot" state

tserong:
Related issue: https://github.com/harvester/harvester/issues/7457

TBH I found this extremely difficult to document clearly, so I hope it makes sense. Also, I'm just about to do another test upgrade following the instructions exactly as written here to make sure I didn't miss anything... Will update when done.

### DIFF
--- a/docs/upgrade/v1-4-0-to-v1-4-1.md
+++ b/docs/upgrade/v1-4-0-to-v1-4-1.md
@@ -14,6 +14,24 @@ An **Upgrade** button appears on the **Dashboard** screen whenever a new Harvest
 
 For air-gapped environments, see [Prepare an air-gapped upgrade](./automatic.md#prepare-an-air-gapped-upgrade).
 
+:::info important
+
+Before upgrading from v1.4.0 to v1.4.1, check the current disk usage of the OS images on each node. To do this, log in via `ssh` and run the following command:
+
+```
+# du -sh /run/initramfs/cos-state/cOS/*
+1.7G    /run/initramfs/cos-state/cOS/active.img
+3.1G    /run/initramfs/cos-state/cOS/passive.img
+```
+If `passive.img` is using 3.1G of disk space as in the above example, become root and run the following commands:
+```
+# mount -o remount,rw /run/initramfs/cos-state
+# fallocate --dig-holes /run/initramfs/cos-state/cOS/passive.img
+# mount -o remount,ro /run/initramfs/cos-state
+```
+This will turn `passive.img` into a sparse file, which should then only be using 1.7G of disk space, the same as `active.img`. Doing this ensures that there will be enough free space on the COS_STATE partition to avoid hitting the [Upgrade is stuck in the "Waiting Reboot" state](#3-upgrade-is-stuck-in-the-waiting-reboot-state) problem.
+:::
+
 
 ### Update Harvester UI Extension on Rancher v2.10.1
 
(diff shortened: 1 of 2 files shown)


tserong:
Ugh, hang on, this bit is wrong:

```
# du -sh /run/initramfs/cos-state/cOS/*
1.7G    /run/initramfs/cos-state/cOS/active.img
3.1G    /run/initramfs/cos-state/cOS/passive.img
```

That will only be the case _before_ the upgrade is attempted. I'll need to rework that bit.

tserong:
OK, that's better...

tserong:
Actually, I should probably add an "important" note to the General Information section saying to sparsify passive.img _before_ starting the upgrade to avoid the issue entirely.

ibrokethecloud:
should we change the wording to ask users to check partition sizes before triggering an upgrade in case they have used an alternate data disk?

converting the passive image to a sparse file will avoid the issue all together and result in a smoother upgrade experience?

tserong:
> should we change the wording to ask users to check partition sizes before triggering an upgrade in case they have used an alternate data disk?
> 
> converting the passive image to a sparse file will avoid the issue all together and result in a smoother upgrade experience?

Yeah, that's what I was trying to do with the "important note" up the top of the page.

docs/upgrade/v1-4-0-to-v1-4-1.md
```diff
@@ -144,3 +162,117 @@ You can perform any of the following workarounds:
...
+
+To fix this problem:
+
+1. If you have two or more nodes, on each of the remaining nodes that are yet to be upgraded, log in via `ssh` and check the disk usage of `active.img` and `passive.img`:
+   ```
+   # du -sh /run/initramfs/cos-state/cOS/*
+   1.7G    /run/initramfs/cos-state/cOS/active.img
+   3.1G    /run/initramfs/cos-state/cOS/passive.img
+   ```
+   If `passive.img` is using 3.1G of disk space, become root and run the following commands:
```
jillian-maroket, line 227:
```suggestion
   If `passive.img` consumes 3.1G of disk space, run the following commands using the root account:
```

docs/upgrade/v1-4-0-to-v1-4-1.md
```diff
@@ -144,3 +162,117 @@ You can perform any of the following workarounds:
...
+   ```
+   # mount -o remount,rw /run/initramfs/cos-state
+   # fallocate --dig-holes /run/initramfs/cos-state/cOS/passive.img 
+   # mount -o remount,ro /run/initramfs/cos-state
+   ```
+   This will turn `passive.img` into a sparse file, which should then only be using 1.7G of disk space, the same as `active.img`. Doing this ensures that there will be enough free space on those other nodes to not get stuck again as the upgrade proceeds.
+
+2. On the stuck node, log in via `ssh`, become root and run the following commands:
```
jillian-maroket, line 235:
```suggestion
   `passive.img` is converted to a sparse file, which should only consume 1.7G of disk space (the same as `active.img`). This ensures that the other nodes have enough free space, preventing the upgrade process from becoming stuck again.

1. Access the stuck node via SSH, and then run the following commands using the root account:
```

docs/upgrade/v1-4-0-to-v1-4-1.md
```diff
@@ -144,3 +162,117 @@ You can perform any of the following workarounds:
...
+   # tune2fs -L COS_ACTIVE /run/initramfs/cos-state/cOS/active.img
+   # mount -o remount,ro /run/initramfs/cos-state
+   ```
+   This will copy your existing (clean) `passive.img` over the corrupt `active.img`, and set the label correctly so that it can be used.
+
+3. Reboot the stuck node. When you get to the GRUB boot screen, it will still default to "Harvester v1.4.1 (fallback)". Select the first entry instead ("Harvester v1.4.1"). Despite the version shown on the GRUB boot screen, this will actually now boot the system into Harvester v1.4.0.
+
+4. Copy `rootfs.squashfs` from the Harvester v1.4.1 ISO image to somewhere convenient on the stuck node. You can either mount the ISO directly on the stuck node, or you can mount it on another system and copy the file over via `scp`.
+
+5. Log in to the stuck node via `ssh`, become root, and run the following commands (replace `rootfs.squashfs` on the fourth line with whatever path you copied that file to):
```
jillian-maroket, line 249:
```suggestion
    The existing (clean) `passive.img` is copied over the corrupted `active.img` and the label is set correctly.

1. Reboot the stuck node, and then select the first entry ("Harvester v1.4.1") on the GRUB boot screen.

    The GRUB boot screen initially displays "Harvester v1.4.1 (fallback)" by default. Despite the displayed version, the system boots into Harvester v1.4.0.

1. Copy `rootfs.squashfs` from the Harvester v1.4.1 ISO to a convenient location on the stuck node.

    The ISO can be mounted either on the stuck node or on another system. You can copy the file using the `scp` command.

1. Access the stuck node via SSH, and then run the following commands using the root account:
```

jillian-maroket (commented):
Review done. Let me know if any rewrites have caused the meaning to change. Please indent the code blocks in the ordered lists.

tserong:
Thanks @jillian-maroket that's excellent! I've indented the code blocks and sync'd with the versioned docs. The only tiny tweak I made to your changes was in the note up the top where I changed "the other nodes have enough free space" to "each node has enough free space", because here we're doing each node in advance (whereas later when this is discussed in detail in the known issue section, we're only doing this to the "other nodes", i.e. the ones that aren't yet stuck). Hope that makes sense.

jillian-maroket (approved):
LGTM. Thanks!

PR #725, opened 2025-03-11
Add upgrade docs for v1.4.2

tserong:


tserong:
Note: need to double check that the rancher v2.10.1 UI extension instructions and pre-drained known issue are still current.

innobead:
> Note: need to double check that the rancher v2.10.1 UI extension instructions and pre-drained known issue are still current.

What's the issue number for the known issue?

tserong:
Just to add to the previous comment:
- I've kept "Update Harvester UI Extension on Rancher v2.10.1" in the v1.4.1->v1.4.2 notes because it says "This issue, which exists in Rancher v2.10.0 and v2.10.1, will be fixed in v2.10.2."
- I've kept "Upgrade is stuck in the "Pre-drained" state" because that was a known issue in v1.4.0->v1.4.1, and https://github.com/harvester/harvester/issues/7325 which should fix it on the Harvester side is scheduled for v1.6.0. Based on that, the issue should still be mentioned for v1.4.2.

bk201 (approved):

jillian-maroket (approved):
Thanks!

PR #726, opened 2025-03-11
fix: Fixed typo in harvester configuration example

ihcsim:
This PR fixed a typo found in the Harvester configuration example where the cluster DNS field should be `cluster_dns`, not `cluster_dns_ip`. This is a 1.5 feature.

Relevant issue: https://github.com/harvester/harvester/issues/7797


jillian-maroket (approved):
Thanks!

PR #727, opened 2025-03-14
Mention LHv2 doesn't support Storage Network

tserong:
Related issue: https://github.com/harvester/harvester/issues/7830

### DIFF
--- a/docs/advanced/longhorn-v2.md
+++ b/docs/advanced/longhorn-v2.md
@@ -34,6 +34,7 @@ The Longhorn V2 Data Engine currently does not support the following operations:
 
 - Backing image creation and usage
 - Live migration
+- Storage network
 - Volume cloning
 - Volume encryption
 - Volume expansion
--- a/docs/advanced/storagenetwork.md
+++ b/docs/advanced/storagenetwork.md
@@ -14,7 +14,8 @@ For more information, please see [Longhorn Storage Network](https://longhorn.io/
 
 :::note
 
-Configuring Longhorn settings directly is not recommended, as this can lead to untested situations.
+- Avoid configuring Longhorn settings directly, as this can result in unexpected or unwanted system behavior.
+- You can only use the Longhorn Storage Network with the default Longhorn V1 Data Engine. Usage with the V2 Data Engine is not yet supported.
 
 :::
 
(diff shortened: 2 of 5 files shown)


docs/advanced/longhorn-v2.md
```diff
@@ -34,6 +34,7 @@ The Longhorn V2 Data Engine currently does not support the following operations:
 
 - Backing image creation and usage
```
innobead, line 35: We can remove this, since we adopt PVC for image in 3rd party storage for v2.

innobead (commented):
LGTM. Just one feedback.

docs/advanced/longhorn-v2.md
tserong: Good catch, although it might be worth doing that in a separate PR, because I think there's other things further down that will also need changing for v1.5 using 3rd party store (e.g. the current limitation on boot disks being V1 only)

docs/advanced/storagenetwork.md
```diff
@@ -10,11 +10,12 @@ title: "Storage Network"
...
 Harvester uses Longhorn as its built-in storage system to provide block device volumes for VMs and Pods. If the user wishes to isolate Longhorn replication traffic from the Kubernetes cluster network (i.e. the management network) or other cluster-wide workloads. Users can allocate a dedicated storage network for Longhorn replication traffic to get better network bandwidth and performance.
 
-For more information, please see [Longhorn Storage Network](https://longhorn.io/docs/1.4.3/advanced-resources/deploy/storage-network/)
+For more information, please see [Longhorn Storage Network](https://longhorn.io/docs/1.8.1/advanced-resources/deploy/storage-network/)
 
 :::note
 
-Configuring Longhorn settings directly is not recommended, as this can lead to untested situations.
+- Configuring Longhorn settings directly is not recommended, as this can lead to untested situations.
+- The Longhorn Storage network is only supported when using the default Longhorn V1 Data Engine. It is not yet supported when using the V2 Data Engine.
```
jillian-maroket, line 18:
```suggestion
- Avoid configuring Longhorn settings directly, as this can result in unexpected or unwanted system behavior.
- You can only use the Longhorn Storage Network with the default Longhorn V1 Data Engine. Usage with the V2 Data Engine is not yet supported.
```

jillian-maroket (commented):
I suggested a few minor changes.

tserong:
Thanks @jillian-maroket !

Vicente-Cheng (approved):
thanks!

jillian-maroket (approved):

PR #732, opened 2025-03-20
docs: use template to describe the uiplugin endpoint

Yu-Jack:
https://github.com/harvester/harvester/issues/7353#issuecomment-2739170632

### DIFF
--- a/docs/airgap.md
+++ b/docs/airgap.md
@@ -169,4 +169,8 @@ spec:
     endpoint: http://ui-plugin-catalog-svc.cattle-ui-plugin-system:8080/plugin/harvester-1.0.3
 ```
 
-Ensure that `svc.namespace` is accessible from Rancher. If that endpoint is not accessible, you can directly use a cluster IP such as `10.43.33.58:8080/plugin/harvester-1.0.3`.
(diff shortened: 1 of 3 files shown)


docs/airgap.md
```diff
@@ -169,4 +169,8 @@ spec:
     endpoint: http://ui-plugin-catalog-svc.cattle-ui-plugin-system:8080/plugin/harvester-1.0.3
 ```
 
-Ensure that `svc.namespace` is accessible from Rancher. If that endpoint is not accessible, you can directly use a cluster IP such as `10.43.33.58:8080/plugin/harvester-1.0.3`.
\ No newline at end of file
+Ensure that `{image name}-svc.namespace` is accessible from Rancher. If that endpoint is not accessible, you can directly use a cluster IP such as `10.43.33.58:8080/plugin/harvester-1.0.3`.
+
+The endpoint is built from the imported image name without organization name. For example:
+  - Wtih `rancher/harvester-ui-catalog` image, it'll be is `harvester-ui-catalog-svc`.
```
a110605, line 175: typo `Wtih`

Yu-Jack:
no need

PR #747, opened 2025-04-08
Add note about imported VMs and third-party storage

votdev:
- Fix typos.
- Rephrase several sentences. E.g. remove the Harvester version because it is already mentioned in the note above the paragraph.
- Add note about the usage of `StorageClass`.
- Replace `VMWare` by `VMware`.

Related to: https://github.com/harvester/harvester/issues/7843


starbops (approved):
LGTM, thank you.

PR #749, opened 2025-04-09
doc(ui-ext): update harvester-ui-extension v1.0.5 support matrix table

a110605:
Harvester-ui-extension v1.0.5 is released to gh-pages. https://github.com/harvester/harvester-ui-extension/releases

Add v1.0.5 mapping table.


starbops (approved):
LGTM, thank you.

PR #756, opened 2025-04-17
Update Longhorn V2 Data Engine docs for Harvester v1.5.0

tserong:
#### Related Issue(s):
https://github.com/harvester/harvester/issues/7521
https://github.com/harvester/harvester/issues/7522

### DIFF
--- a/docs/advanced/longhorn-v2.md
+++ b/docs/advanced/longhorn-v2.md
@@ -7,7 +7,7 @@ Description: How to enable and use the Longhorn V2 Data Engine
 ---
 
 <head>
-  <link rel="canonical" href="https://docs.harvesterhci.io/v1.4/advanced/longhorn-v2"/>
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.5/advanced/longhorn-v2"/>
 </head>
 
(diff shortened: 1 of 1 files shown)


tserong:
Upgrades from Harvester v1.4.x (LH v1.7.x) to Harvester v1.5.x (LH v1.8.x) with LHv2 volumes can be somewhat problematic in my testing:

- If you're using the AIO bdev driver, any existing volumes just flat out don't work after the upgrade (see https://github.com/longhorn/longhorn/issues/10461)
- If you're using the NVMe bdev driver, existing volumes work after the upgrade but remain stuck using the v1.7.x longhorn engine image and AFAICT there's no way to upgrade that inplace.
- Any migration effort for existing volumes results in two copies of the entire volume data and if it was sparse originally, it won't be after the copy.
- TL;DR: IMO you're really better off just trashing existing pre-upgrade V2 volumes and starting again post-upgrade.

I've tried to cover this in docs/advanced/longhorn-v2.md, but we may also want to call it out in the v1.4.2->v1.5.0 upgrade page once that's in place.

Vicente-Cheng (approved):
lgtm, thanks!

jillian-maroket:
In the interest of time, I suggest that we merge as is. I'll just create another PR for minor changes after the release.

WebberHuang1118 (approved):
LGTM, thanks.

Vicente-Cheng:
Thanks @jillian-maroket, 
Since @tserong is on vacation, I'll merge it as you suggested.

Please help create another PR for minor changes.

PR #757, opened 2025-04-21
docs: add information for Harvester#8096 i.e. Longhorn#10741

WebberHuang1118:
#### Problem:
Can NOT delete an oversized Not Ready volume

#### Solution:
As a temporary workaround for Harvester `v1.5.0`. The underlying Longhorn issue has been addressed and the fix will be included in Harvester `v1.5.1` (LH `v1.8.2`).

#### Related Issue(s):
https://github.com/longhorn/longhorn/issues/10741
https://github.com/harvester/harvester/issues/8096

### DIFF
--- a/docs/volume/create-volume.md
+++ b/docs/volume/create-volume.md
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
 
 ![create-empty-volume](/img/v1.2/volume/create-empty-volume.png)
 
+### Oversized Volumes
+
+In Harvester v1.5.0, which uses Longhorn v1.8.1, oversized volumes (for example, 999999 Gi in size) are marked **Not Ready** and cannot be deleted.
+
+To resolve this issue, perform the following steps:
+
+1. Temporarily remove the PVC webhook rule.
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p='[{"op": "remove", "path": "/webhooks/0/rules/17"}]'
+   ```
+
+1. Wait for the related PVC to be deleted.
+
+1. Restore the PVC webhook rule to re-enable validation.
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p='[{"op": "add", "path": "/webhooks/0/rules/-", "value": {"apiGroups":[""],"apiVersions":["v1"],"operations":["UPDATE"],"resources":["persistentvolumeclaims"],"scope":"Namespaced"}}]'
+   ```
+
+The issue will be addressed in Longhorn v1.8.2, which will likely be included in Harvester v1.5.1.
+
+Related issues:
+- Harvester: [Issue #8096](https://github.com/harvester/harvester/issues/8096)
+- Longhorn: [Issue #10741](https://github.com/longhorn/longhorn/issues/10741)
+
 </TabItem>
 <TabItem value="api" label="API">
 


docs/volume/create-volume.md
```diff
@@ -29,6 +29,44 @@ description: Create a volume from the Volume page.
...
+   ```
+
+2. Wait for the related PVC to be deleted.
+
+3. Restore the PVC update rule to re-enable validation:
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p='[{"op": "add", "path": "/webhooks/0/rules/-", "value": {"apiGroups":[""],"apiVersions":["v1"],"operations":["UPDATE"],"resources":["persistentvolumeclaims"],"scope":"Namespaced"}}]'
```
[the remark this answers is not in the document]
WebberHuang1118: Checked the related command can't be applied

docs/volume/create-volume.md
```diff
@@ -29,6 +29,44 @@ description: Create a volume from the Volume page.
...
+   ```
+
+To prevent this issue, you can remove the PVC webhook rule before performing the operation:
```
Vicente-Cheng, line 60: I am not sure about this hint.
We should not always disable the PVC webhook if we want it to perform other validations for PVC.
WDYT?

docs/volume/create-volume.md
```diff
@@ -29,6 +29,44 @@ description: Create a volume from the Volume page.
 
 ![create-empty-volume](/img/v1.2/volume/create-empty-volume.png)
 
+### Oversized Volume
+
+:::caution
+
+Creating an oversized volume (e.g., 999999G) can cause the volume to be marked as "Not Ready" and undeletable. This issue has been observed in Harvester version `v1.5.0` (Longhorn version `v1.8.1`).
```
starbops, line 36: Do we have a clear number for users to know what volume size will cause an issue?

WebberHuang1118: Theoretically, `resource.Quantity` values smaller than `2^63 - 1` (i.e., the maximum for an int64) should not require a pointer. 

However, in practice, this depends on the Kubernetes implementation. For example, even a value like `999999GB` may still result in a `resource.Quantity` that contains a pointer, which can cause direct comparisons to fail.

docs/volume/create-volume.md
WebberHuang1118: Yes, good point, thanks. I will remove this part.

Vicente-Cheng (dismissed):
thanks for the quick update

docs/volume/create-volume.md
starbops: Understood. Since we only have a post-reaction as the workaround, I'm okay with not knowing the limitation beforehand.

starbops (approved):
LGTM, thank you!

docs/volume/create-volume.md
```diff
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
...
```
jillian-maroket, line 36:
```suggestion
### Oversized Volumes

In Harvester v1.5.0, which uses Longhorn v1.8.1, oversized volumes (for example, 999999 Gi in size) are marked **Not Ready** and cannot be deleted.
```

jillian-maroket: The text block is too long to be placed in a note. I think the heading is sufficient for drawing attention to the text.

docs/volume/create-volume.md
```diff
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
...
+
+This issue is related to the [Longhorn issue #10741](https://github.com/longhorn/longhorn/issues/10741) and the [Harvester issue #8096](https://github.com/harvester/harvester/issues/8096). It will be addressed in Longhorn version `v1.8.2`, which is expected to be included in Harvester version `v1.5.1`.
```
jillian-maroket, line 38: Please remove this. I placed the info at the bottom of the section.

docs/volume/create-volume.md
```diff
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
...
+
+To resolve this issue, follow these steps:
+
+1. Temporarily remove the PVC webhook rule:
```
jillian-maroket, line 42:
```suggestion
To resolve this issue, perform the following steps:

1. Temporarily remove the PVC webhook rule.
```

docs/volume/create-volume.md
```diff
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
...
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p='[{"op": "remove", "path": "/webhooks/0/rules/17"}]'
+   ```
+
+2. Wait for the related PVC to be deleted.
+
+3. Restore the PVC update rule to re-enable validation:
```
jillian-maroket, line 52:
```suggestion
1. Wait for the related PVC to be deleted.

1. Restore the PVC webhook rule to re-enable validation.
```

jillian-maroket: I changed `update rule` to `webhook rule` because that's what the user supposedly removed in step 1. If I misunderstood, please change the text.

docs/volume/create-volume.md
```diff
@@ -29,6 +29,36 @@ description: Create a volume from the Volume page.
...
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p='[{"op": "add", "path": "/webhooks/0/rules/-", "value": {"apiGroups":[""],"apiVersions":["v1"],"operations":["UPDATE"],"resources":["persistentvolumeclaims"],"scope":"Namespaced"}}]'
+   ```
+
+:::
```
jillian-maroket, line 60:
```suggestion
The issue will be addressed in Longhorn v1.8.2, which will likely be included in Harvester v1.5.1.

Related issues:
- Harvester: [Issue #8096](https://github.com/harvester/harvester/issues/8096)
- Longhorn: [Issue #10741](https://github.com/longhorn/longhorn/issues/10741)
```

jillian-maroket (commented):
Review done

docs/volume/create-volume.md
WebberHuang1118: updated per suggesion, thanks.

jillian-maroket (approved):

PR #760, opened 2025-04-22
Create v1.6.0 doc and update config files

jillian-maroket:
Issue: [7756](https://github.com/harvester/harvester/issues/7756) Release v1.5.0

Changes: I followed the procedures in [Publish a New Doc Version](https://github.com/harvester/docs/wiki/Publish-a-New-Doc-Version), which I based on the v1.4.0 release workflow. A few steps are missing, so I'll update the wiki once the doc is published. 

*Note: The earlier commits resulted in errors because `version-v1.5` was missing. I could not create it before modifying the config files because some doc PRs were still pending.*


bk201 (approved):

starbops (approved):
LGTM, thank you.

PR #774, opened 2025-05-06
docs: Adjust Settings Value Examples

irishgordo:
* adjust settings value examples, in YAML wrapping {} JSON blobs will need to be done with something like single-quotes '', and have other content escaped where applicable

Resolves: docs/adj-settings
See also: https://github.com/harvester/harvester/issues/7456

#### Problem:
JSON values need to be wrapped and escaped

#### Solution:
Updating examples

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7456

#### Test plan:
N/A

https://github.com/harvester/harvester/issues/7456


Yu-Jack:
Hi @irishgordo FYI, I just rebased this PR for testing preview.

PR #776, opened 2025-05-08
ci: fix netlify preview args

Yu-Jack:
quick fix


Yu-Jack:
We could ignore the checks inside this PR because the workflow hasn't been updated yet (https://github.com/harvester/docs/actions/runs/14897091663/job/41842275315).

Vicente-Cheng (approved):

PR #778, opened 2025-05-08
Update upgrade matrix and known issues for v1.4.3 release

tserong:
#### Problem:
- Need to update the upgrade matrix to reflect that it's possible to go:
  - v1.4.2/v1.4.3 --> v1.5.0
  - v1.4.1/v1.4.2 --> v1.4.3
- As v1.4.3 adds Longhorn v1.7.3, we have a new known issue to document (https://github.com/harvester/harvester/issues/8096)

#### Solution:
This PR, but note that because v1.4.3 is not yet released, I have _not_ added the v1.4.3 upgrade docs to the versioned_docs yet. Once everyone's cool with the content here, I'll update the PR to add changes to versioned_docs and merge when v1.4.3 goes out.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8193
https://github.com/harvester/harvester/issues/8096

### DIFF
--- a/docs/upgrade/v1-4-1-to-v1-4-3.md
+++ b/docs/upgrade/v1-4-1-to-v1-4-3.md
@@ -12,6 +12,8 @@ title: "Upgrade from v1.4.1/v1.4.2 to v1.4.3"
 
 An **Upgrade** button appears on the **Dashboard** screen whenever a new Harvester version that you can upgrade to becomes available. For more information, see [Start an upgrade](./automatic.md#start-an-upgrade).
 
+Harvester v1.4.2 and v1.4.3 use the same minor version of RKE2 (v1.31). This makes it possible to upgrade directly from Harvester v1.4.1 to v1.4.3, without needing to upgrade to v1.4.2 first.
+
 For air-gapped environments, see [Prepare an air-gapped upgrade](./automatic.md#prepare-an-air-gapped-upgrade).
 
 ## Known issues
--- a/docs/upgrade/v1-4-1-to-v1-4-3.md
+++ b/docs/upgrade/v1-4-1-to-v1-4-3.md
@@ -1,11 +1,11 @@
 ---
 sidebar_position: 3
-sidebar_label: Upgrade from v1.4.2 to v1.4.3
-title: "Upgrade from v1.4.2 to v1.4.3"
+sidebar_label: Upgrade from v1.4.1/v1.4.2 to v1.4.3
+title: "Upgrade from v1.4.1/v1.4.2 to v1.4.3"
 ---
 
 <head>
-  <link rel="canonical" href="https://docs.harvesterhci.io/v1.5/upgrade/v1-4-2-to-v1-4-3"/>
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.5/upgrade/v1-4-1-to-v1-4-3"/>
 </head>
 
 ## General information
--- a/docs/upgrade/v1-4-1-to-v1-4-3.md
+++ b/docs/upgrade/v1-4-1-to-v1-4-3.md
@@ -108,3 +108,38 @@ To fix the issue, perform any of the following actions:
 
 Related issues:
 - [[BUG] AirGap Upgrades Seem Blocked with Fluentbit/FluentD](https://github.com/harvester/harvester/issues/7955)
+
+### 2. Oversized Volumes
+
+In Harvester v1.4.3, which uses Longhorn v1.7.3, oversized volumes (for example, 999999 Gi in size) are marked **Not Ready** and cannot be deleted.
+
+To resolve this issue, perform the following steps:
+
+1. Temporarily remove the PVC webhook rule.
+
+   ```bash
+   RULE_INDEX=$(kubectl get \
+     validatingwebhookconfiguration longhorn-webhook-validator -o json \
+     | jq '.webhooks[0].rules | map(.resources[0] == "persistentvolumeclaims") | index(true)')
+   
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
+     -p="[{'op': 'remove', 'path': '/webhooks/0/rules/$RULE_INDEX'}]"
+   ```
+
+1. Wait for the related PVC to be deleted.
+
+1. Restore the PVC webhook rule to re-enable validation.
+
+   ```bash
+   kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+     --type='json' \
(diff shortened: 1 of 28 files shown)


versioned_docs/version-v1.5/volume/create-volume.md
```diff
@@ -38,9 +38,13 @@ To resolve this issue, perform the following steps:
 1. Temporarily remove the PVC webhook rule.
 
    ```bash
+   RULE_INDEX=$(kubectl get \
```
w13915984028, line 41: If any failure happens, the `RULE_INDEX` is set to 0 and might remove 0 accidentally

docs/upgrade/v1-4-1-to-v1-4-3.md
```diff
@@ -1,11 +1,11 @@
 ---
 sidebar_position: 3
-sidebar_label: Upgrade from v1.4.2 to v1.4.3
-title: "Upgrade from v1.4.2 to v1.4.3"
+sidebar_label: Upgrade from v1.4.1/v1.4.2 to v1.4.3
+title: "Upgrade from v1.4.1/v1.4.2 to v1.4.3"
```
w13915984028, line 4: is v141->v143 direct upgrade allowed?

add an explanation in the md that the RKE2/K8S version does not change from v142 to v143, thus v141 can upgrade to v143 directly

w13915984028 (dismissed):
LGTM, thanks.

docs/upgrade/v1-4-1-to-v1-4-3.md
```diff
@@ -108,3 +108,38 @@ To fix the issue, perform any of the following actions:
 
 Related issues:
 - [[BUG] AirGap Upgrades Seem Blocked with Fluentbit/FluentD](https://github.com/harvester/harvester/issues/7955)
+
+### 2. Oversized Volumes
```
w13915984028, line 112: Btw, normally the known issues & workarounds are placed in troubleshooting https://docs.harvesterhci.io/v1.4/troubleshooting/index, and put a link in the normal document to make it clean.

But this is also not fully followed through all the chapters.

@jillian-maroket WDYT? thanks.

versioned_docs/version-v1.5/volume/create-volume.md
tserong: Good point, thank you. I did a little further testing, if the rule can't be found (e.g. if it's already been removed), I end up with `RULE_INDEX=null`:

```
# RULE_INDEX=$(kubectl get \
    validatingwebhookconfiguration longhorn-webhook-validator -o json \
    | jq '.webhooks[0].rules | map(.resources[0] == "persistentvolumeclaims") | index(true)')

# echo $RULE_INDEX
null
```

If something else is broken, `RULE_INDEX` is set to an empty string:

```
# RULE_INDEX=$(KUBECONFIG=BOGUS kubectl get \
    validatingwebhookconfiguration longhorn-webhook-validator -o json \
    | jq '.webhooks[0].rules | map(.resources[0] == "persistentvolumeclaims") | index(true)')
# echo $RULE_INDEX

```

In both cases (`RULE_INDEX=null` or `RULE_INDEX=`), if I try the following patch operation it fails without making any changes:

```
# kubectl patch validatingwebhookconfiguration longhorn-webhook-validator      --type='json'      -p="[{'op': 'remove', 'path': '/webhooks/0/rules/$RULE_INDEX'}]"
The request is invalid: the server rejected our request due to an error in our request
```

So I think it's safe as-is. But, to be clear for the user, I could add an extra step to check that RULE_INDEX is valid. Maybe something like this:

```
RULE_INDEX=$(kubectl get validatingwebhookconfiguration [...])

# verify RULE_INDEX is numeric -- the following should print a number, e.g. 14 or 17
echo $RULE_INDEX

kubectl patch validatingwebhookconfiguration [....]
```

WDYT?

docs/upgrade/v1-4-1-to-v1-4-3.md
tserong: Yes, upgrade from v1.4.1 is allowed because we're only doing very minor component bumps between v1.4.2 and v1.4.3:

- RKE2 v1.31.4 --> v1.31.7
- Longhorn v1.7.2 --> LH v1.7.3

I'm happy to add a note to clarify this of course.

docs/upgrade/v1-4-1-to-v1-4-3.md
tserong: TBH I wasn't sure where best to put this. It was already mentioned in Volume Management: Create a Volume in the v1.5 docs, so I added it in the same place in the v1.4 docs, but I figured someone upgrading from v1.4.1 or v1.4.2 wouldn't necessarily think to go looking there, so I added it to the upgrade page as well. I'll defer to @jillian-maroket for what the best choice is.

versioned_docs/version-v1.5/volume/create-volume.md
w13915984028: we may organize the code into a function, and use  `if  [[ -z $RULE_INDEX ]] ` to print message or delete the object

docs/upgrade/v1-4-1-to-v1-4-3.md
jillian-maroket: When I joined the team, each upgrade page already had a section for known issues and engineers just kept updating those files. I'm fine with continuing this practice for now. We can discuss it in one of the weekly meetings.

jillian-maroket (approved):

docs/upgrade/v1-4-1-to-v1-4-3.md
w13915984028: Sorry I put the wrong link.

There is an `oversized volumes` section in `versioned_docs/version-v1.5/volume/create-volume.md`, which can simply link to the known issues in `...upgrade.md`, to keep the `create-volume.md` clean.

Vicente-Cheng (approved):
thanks!

starbops (dismissed):
LGTM, thank you.

WebberHuang1118 (dismissed):
LGTM, thanks.

tserong:
I've made two small changes to address a couple of earlier review comments:

- A note about RKE2 versions
- A check that $RULE_INDEX is valid

Everything else remains the same. Can you please re-check when you get a chance @w13915984028 and @jillian-maroket?

w13915984028 (dismissed):
LGTM, thanks.

jillian-maroket (approved):
Thanks!

ihcsim (approved):

jillian-maroket (approved):

starbops (approved):

PR #790, opened 2025-05-26
docs: Add more details about http-proxy

vatsalparekh:
#### Problem:

https://github.com/harvester/harvester/issues/8337

Improving http proxy related documentation

### DIFF
--- a/docs/advanced/settings.md
+++ b/docs/advanced/settings.md
@@ -191,9 +191,12 @@ Harvester sends a graceful shutdown signal to any VM that is stopped using the H
 
 **Definition**: HTTP proxy used to access external services, including downloading of images and backup to S3 services.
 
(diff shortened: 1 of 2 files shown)


jillian-maroket:
@starbops PTAL. Thanks!

starbops (commented):
Thanks for the PR @vatsalparekh. In general, the aspect we want to improve is that users should be able to determine precisely which components, such as pods and system services, will honor these HTTP proxy-related settings by reading through these specific paragraphs. So, it'd be great if you can list down all the components affected.

To achieve this, you might need to trace the source code, as we use a mutator to inject an HTTP proxy-related environment variable into only a set of pods.

Additionally, there is a specific case that we would like to address, as described in the original issue. The key is to set HTTP proxy-related settings when importing the Harvester cluster to the external Rancher Manager. Users typically don't need to modify the Rancher Manager's HTTP proxy setting. These are two different things. It'd be great if we could provide links to the relevant parts in other projects' documentation.

akashraj4261 (approved):

PR #798, opened 2025-06-09
Add known issue for RWX non-root user permission issues in v1.4.3

tserong:
Note that here I'm deliberately suggesting the user directly edit the longhorn-manager daemonset.  This means that if the user attempts to later upgrade to v1.5.x _without_ reverting the change, upgrade will be prevented with the following error:

```
  admission webhook "validator.harvesterhci.io" denied the request:
  managed chart harvester is not ready, please wait for it to be ready
```

This is a safety measure, to make really sure the hotfix image is removed before subsequent upgrade, because if we _don't_ remove it (e.g. if we made a nice clean patch to the managedchart itself to apply the hotfix image, which would avoid the above error), then proceeded to upgrade to Harvester v1.5.x, then all of Longhorn would be updated to v1.8.x _except_ for the share manager, which would remain using the old v1.7.3-hotfix-1 image!

I figured it would be too confusing to add all that detail to the docs, so instead have recorded it here for posterity.

Related issue: https://github.com/harvester/harvester/issues/8354

### DIFF
--- a/docs/upgrade/v1-4-1-to-v1-4-3.md
+++ b/docs/upgrade/v1-4-1-to-v1-4-3.md
@@ -147,3 +147,56 @@ Related issues:
 - Harvester: [Issue #8096](https://github.com/harvester/harvester/issues/8096)
 - Longhorn: [Issue #10741](https://github.com/longhorn/longhorn/issues/10741)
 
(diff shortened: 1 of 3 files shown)


starbops (approved):
Technically LGTM, thank you.

jillian-maroket (approved):
LGTM. Thanks!

PR #800, opened 2025-06-17
fix: window path

Yu-Jack:
#### Problem:

Real window path is different from documentation.

#### Solution:

Fix window path.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7594


jillian-maroket (approved):

albinsun (approved):
lgtm

PR #809, opened 2025-06-30
chore: fix puppeteer building issue with PUPPETEER_DOWNLOAD_BASE_URL

Yu-Jack:
It seems it's a puppeteer issue. I tried in my personal repository, and it works. 

image

Since we're using pull_request_target, so we must merge this PR, then we're able to test it.

Reference: https://github.com/puppeteer/puppeteer/issues/12833


jillian-maroket (approved):
Thanks!

PR #814, opened 2025-07-04
Add information about v1.5.0 to v1.5.1 upgrade path

jillian-maroket:
Related issue: [8432](https://github.com/harvester/harvester/issues/8342)

Changes:
- Updated upgrade paths table
- Created `docs/upgrade/v1-5-0-to-v1-5-1.md`
- Updated `sidebar_position` value of other upgrade files


starbops (commented):
The first known issue has been fixed in v1.5.0 so we can drop that for the v1.5.0 to v1.5.1 upgrade path.

https://github.com/harvester/harvester/issues/7955#issuecomment-2796087579

jillian-maroket:
> The first known issue (Air-gapped Upgrade Stuck with `ImagePullBackOff` Error in Fluentd and Fluent Bit Pods) has been fixed in v1.5.0 so we can drop that for the v1.5.0 to v1.5.1 upgrade path.
> 
> [harvester/harvester#7955 (comment)](https://github.com/harvester/harvester/issues/7955#issuecomment-2796087579)

@starbops Thank you for reviewing. I have removed the first known issue.

starbops (approved):
LGTM, thank you.

PR #818, opened 2025-07-08
fix: add API Navbar back

Yu-Jack:
The reasons why it showed API sidebar and failed to show API Navbar with correct page are:

- We ignored the `_category_.json`. It showedAPI sidebar.
  https://github.com/harvester/docs/pull/545 

- We added `ignore-openapi` when publishing the document. It deleted page `category/api`.
  https://github.com/harvester/docs/pull/579 

-  We removed API Navbar in `docusaurus.config.js`. It removed API Navbar.
  https://github.com/harvester/docs/pull/634 

Here https://github.com/harvester/docs/pull/388 is how they did to show API Navbar and hide API sidebar before. 

---

Hence, I just revert all the changes to make it available again. Besides, I also remove `ignore-openapi` script because we don't need to when releasing at the first version. Then, I think we also don't need it in pull request review.


Yu-Jack:
Yeah. It works https://686cf716b96ac0061f56f8fd--harvester-preview.netlify.app/v1.5/category/api/.

innobead (approved):

m-ildefons (approved):
Cool thank you!
LGTM

PR #827, opened 2025-07-25
Explain the Automatically Applied Affinity Rules

w13915984028:
#### Problem:

Some affinity rules are applied to vm definition automatically, but not documented.

#### Solution:

Expose them on doc.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8350

#### Test plan:

not required.


w13915984028:
@ihcsim Thanks. 

I have addressed the comments, and added a new section on vm troubleshooting illustrates the unscheduable VM. Please take a new look.

docs/troubleshooting/vm.md
```diff
@@ -454,3 +454,76 @@ After deleting the directory, you must restart the virtual machine so that cloud
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: network.harvesterhci.io/cn2
+                    operator: In
+                    values:
+                      - 'true'
+```
+
+The POD status is `Pending`, the detailed error shows that no node can meet the `Affinity` rule.
```
jillian-maroket, line 484:
```suggestion
The pod status is `Pending`, and the error message indicates that no node meets the affinity rule's criteria.
```

docs/troubleshooting/vm.md
```diff
@@ -454,3 +454,76 @@ After deleting the directory, you must restart the virtual machine so that cloud
...
+    reason: Unschedulable
+    status: "False"
+    type: PodScheduled
+...
+
+```
+
+### Root cause
+
+Based on the definition of VM, Harvester [applies some Affinity rules automatically](../vm/create-vm.md#automatically-applied-affinity-rules). In the above example, the VM attaches to cluster network `cn2`, then Harvester applies `Affifnity` rule `network.harvesterhci.io/cn2`. But there is no node meets this rule and hence the VM is not schedulable.
```
jillian-maroket, line 525:
```suggestion
### Root Cause

Harvester [automatically applies affinity rules](../vm/create-vm.md#automatically-applied-affinity-rules) based on how a virtual machine is configured. In the above example, the virtual machine `vm100` attaches to the cluster network `cn2`, and Harvester applies the affinity rule `network.harvesterhci.io/cn2`. However, no nodes meet the rule's criteria, so the virtual machine cannot be scheduled.
```

docs/troubleshooting/vm.md
```diff
@@ -454,3 +454,76 @@ After deleting the directory, you must restart the virtual machine so that cloud
...
+
+### Workaround
+
+Ensure there are active nodes which has already setup the cluster network `cn2`, then the node has the label `network.harvesterhci.io/cn2` and can schedule the pending pod.
```
jillian-maroket, line 529:
```suggestion
### Workaround

Ensure that the cluster network is configured and the required label (for example, `network.harvesterhci.io/cn2`) is applied on active nodes. In addition, ensure that the pending pod can be scheduled on these nodes.
```

jillian-maroket (commented):
Initial review done

@ihcsim Please check if the structure and phrasing work. It wasn't clear if the items in numbered lists actually followed a particular order. Some section titles also seem off.

docs/troubleshooting/vm.md
w13915984028: Need to keep the original description; the `cn2` has been defined, but there is no active node; it is not `cluster network is configured`;  and the label is added by controller automatically, user can't ensure the label.

w13915984028:
@jillian-maroket @ihcsim 

Thanks for the reviewing, the comments are addressed, some descriptions are rephrased to reflect the dependency/order of vm and network. Please take a new look.

innobead (approved):

innobead:
@harvester/network needs to ensure VM migration works when using KubeOVN in a non-mgmt network in the future. Currently, the upcoming 1.6 only supports overlay network running on the mgmt-network, correct? @rrajendran17

rrajendran17:
@innobead yes that is right, currently for v1.6, overlay network is supported only on mgmt network.

Vicente-Cheng:
We still need @jillian-maroket to do the final check. Thanks!

w13915984028:
Thanks for your helpful review. I am merging the PR to move issue forward.

For possible further polishing, let's update them on another related issue https://github.com/harvester/harvester/issues/8823, thanks.

ihcsim:
@w13915984028 @jillian-maroket Sorry for the delay. The troubleshooting guide is great! When we next iterate this section of the doc again, I think the `Related Networking Concepts` and `Related CPU Pinning Concepts` sections can either be moved to their respective feature pages, or be removed. The key is that the UI shows that the VM is in an `Unschedulable` state and the troubleshooting page points out how to diagnose the problem. There could be all sort of reasons why an VM can become `Unschedulable` due to incompatible node taints and pod toleration settings. We can't possibly list them all in the 'Create Virtual Machine' page.

PR #830, opened 2025-07-28
enhance node deletion operations for Keep Backing Image Integrity

WebberHuang1118:
#### Related Issue(s):
https://github.com/harvester/harvester/issues/7846

cc @derekbit


jillian-maroket (commented):
@WebberHuang1118 Did you intentionally update the v1.3 doc? We do not maintain EOL versions.

Screenshot 2025-07-28 at 3 22 40 PM

COLDTURNIP (approved):
LGTM

innobead:
> > @Vicente-Cheng LMK if you want to change the policy of not adding doc content to EOL versions. This was enforced before I joined the company.
> 
> If the issue only occurred in versions with v1.3.x, since v1.3 is EoL, I think we can skip this document update.
> 
> WDYT? @innobead

YES. No need to work on any outdated issues specific to EOL versions.

cc @harvester/dev

PR #839, opened 2025-08-05
Additional Kernel Parameters

ParadoxGuitarist:
#### Problem:
Docs are missing mentions of kernel params that are required to have the same experience as the iso installer over PXE/Netboot. 

#### Solution:
Add more docs!

#### Related Issue(s):
Issue harvester/harvester#8818

What might be really missing out of this is links to where all the `harvester.installer.*` flags are. I have no idea where that is. I found these from various other places/locations. As far as I can tell, these 4 are needed to replicate the same experience booting from the full iso. As I don't have upstream docs for it, you might want to review the descriptions for accuracy. 

I enabled edits for maintainers so ya'll don't have to get me to make changes you need.


docs/install/pxe-boot-install.md
```diff
@@ -446,3 +446,19 @@ Installation is stopped if the hardware checks fail (because the minimum require
...
+
+Specifies the console teletype to activate at boot. 
+
+### `harvester.install.tty=tty1`
+
+Specifies what teletype console to start the installer on.
+
+### `net.ifnames=1`
+
+Enables predictable network interface names in the installer, which will match the system default after harvester is installed. This allows the system to utilize the correct interfaces.
```
votdev, line 464:
```suggestion
Enables predictable network interface names in the installer, which will match the system default after Harvester is installed. This allows the system to utilize the correct interfaces.
```

votdev:
In my opinion, we should add a link to “/install/harvester-configuration” on the PXE page, along with a brief note explaining that further command line options are explained there.

jillian-maroket (approved):

PR #840, opened 2025-08-06
chore: add v1.6.0 ui-ext suppport matrix

a110605:
#### Problem:
chore: add v1.6.0 ui-ext suppport matrix

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8638

#### Test plan:
image


Vicente-Cheng (approved):
thanks!

PR #845, opened 2025-08-08
Enhance cpu pinning doc

brandboat:
see https://github.com/harvester/harvester/pull/8853#pullrequestreview-3099673102

### DIFF
--- a/docs/vm/cpu-pinning.md
+++ b/docs/vm/cpu-pinning.md
@@ -75,6 +75,12 @@ Allow some time for Harvester to apply the corresponding CPU Manager policy.
 
 1. Verify that CPU Manager is enabled on one or more nodes.
 
+  :::note important
+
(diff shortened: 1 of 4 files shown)


ihcsim (approved):

docs/vm/cpu-pinning.md
```diff
@@ -75,6 +75,12 @@ Allow some time for Harvester to apply the corresponding CPU Manager policy.
 
 1. Verify that CPU Manager is enabled on one or more nodes.
 
+  :::note important
+
+  Ensure that at least one node has CPU Manager enabled. Otherwise, the VM will be stuck in the `unschedulable` state after startup. For more information, see [Node Affinity Rules](./create-vm.md#related-cpu-pinning-concepts).
```
ihcsim, line 80: Not directly related to this PR, at one point, we should see if we can find a better home for the "Related CPU Pinning Concepts" and "Related Networking Concepts" sections. From a "creating vm" UX perspective, I don't think those information is necessary. The list might also grow as the system evolves.

brandboat: filed https://github.com/harvester/harvester/issues/8860

PR #846, opened 2025-08-12
update steps to disable kubeovn operator

ibrokethecloud:
#### Problem:

PR udpates the steps to disable kubeovn cleanly

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8852

### DIFF
--- a/docs/advanced/addons/kubeovn-operator.md
+++ b/docs/advanced/addons/kubeovn-operator.md
@@ -91,9 +91,9 @@ spec:
     joinCIDR: 100.64.0.0/16
     pingerExternalAddress: 1.1.1.1
     pingerExternalDomain: google.com.
-    podCIDR: 10.52.0.0/16
-    podGateway: 10.52.0.1
-    serviceCIDR: 10.52.0.1
+    podCIDR: 10.54.0.0/16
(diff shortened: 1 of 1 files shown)


rrajendran17:
is it ok to leave the crds created for kubeovn as it is ?

ibrokethecloud:
the only crd left is the configuration crd which is introduced by kubeovn-operator addon

All other kubeovn crd's will be removed from by the addon when `configuration` object is deleted

jillian-maroket (commented):
I made a few minor suggestions. PTAL

jillian-maroket (approved):

rrajendran17:
few nits:
1.Since the crds are deleted, the ovn-default and join subnet is deleted automatically.We do not need explicit step to delete them
kubectl delete subnets join ovn-default
2.Do we also need to delete the network policies installed by kubeovn ?

ibrokethecloud:
at times the crd deletion gets blocked because webhook is blocking subnet deletion. hence the step to explicitly try and trigger deletion.

We do not need to remove networkpolicies since there is no clear differentiation between calico and kubeovn based network policies.

PR #860, opened 2025-08-27
Add reference that some CPU is reserved for system

martindekov:
Adding reference in the documentation as part of `Host Managament` that certain part of the CPU resources are reserved for system level operations with link to the cpu pinning.

#### Problem:
When we configure CPU in harvester we expect the configured CPU to appear on the host, but it doesn't. The reason is that small part of it is reserved for system processes.

#### Solution:
In the documentation add reference why this happens on non EOL versions - 1.6/1.5/1.4

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8869

#### Test plan:
Ran locally saw info present in all desired versions:
image
image
image

### DIFF
--- a/versioned_docs/version-v1.6/host/host.md
+++ b/versioned_docs/version-v1.6/host/host.md
@@ -19,6 +19,11 @@ Because Harvester is built on top of Kubernetes and uses etcd as its database, t
 
 ![host.png](/img/v1.2/host/host.png)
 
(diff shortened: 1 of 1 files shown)


jillian-maroket (approved):

w13915984028 (approved):
LGTM, thanks.

PR #872, opened 2025-09-03
Suggest a maximum vCPU quantity to allocate

abonillabeeche:
#### Problem:
When VMs are provisioned vCPUs of a higher quantity than available on the hosts, there will be a warning of exceeding the recommended value from the hypervisor.

#### Solution:
Document the suggested maximum per VM vCPUs for performance purposes.

#### Related Issue(s):
Issue https://github.com/harvester/harvester/issues/9024


brandboat (approved):
LGTM, thanks!

jillian-maroket (approved):

PR #876, opened 2025-09-09
Add description of new longhorn storage value

martindekov:
Adding description of the new longhorn default setting: storageReservedPercentageForDefaultDisk

The documentation is for 1.7 only as only on that
version the configuration is present.

#### Problem:
We always use default longhorn percentage reservation value of 30 and this is not configurable as described here:
https://github.com/harvester/harvester/issues/8200

#### Solution:
Making the longhorn storage reservation on default disk configurable

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8200

#### Test plan:
Spin up local harvester doc and this is how it looks like:
image

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -551,9 +551,7 @@ For more information about how to set the correct value, see [Guaranteed Instanc
 
 **Definition**: Percentage of disk space that will not be allocated to the default disk on each new Longhorn node.
 
-Harvester creates a separate partition on the boot disk for the Longhorn default disk. Reserving space on the default root disk for longhorn is redundant in that case.
-
-For more information, see [Storage Reserved Percentage For Default Disk](https://longhorn.io/docs/1.9.1/references/settings/#storage-reserved-percentage-for-default-disk) in the Longhorn documentation.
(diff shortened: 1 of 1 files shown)


w13915984028 (approved):
LGTM, thanks.

docs/install/harvester-configuration.md
```diff
@@ -545,6 +545,29 @@ For more information about how to set the correct value, see [Guaranteed Instanc
         guaranteedInstanceManagerCPU: 6
 ```
 
+### `install.harvester.longhorn.default_settings.storageReservedPercentageForDefaultDisk`
+
+**Versions**: v1.7.0 and later
+
+**Definition**: Percentage of disk space that will not be allocated to the default disk on each new Longhorn node.
+
+Harvester creates a separate partition on the boot disk for the Longhorn default disk. Reserving space on the default root disk for longhorn is redundant in that case.
```
jillian-maroket, line 554:
```suggestion
Harvester creates a dedicated partition on the boot disk for the Longhorn default disk, so you do not need to reserve any space for it on the root disk.
```

jillian-maroket (commented):
Review done

martindekov:
Thanks for the review @jillian-maroket your feedback is addressed please take another look when you can

jillian-maroket (approved):
Thank you. LGTM

PR #881, opened 2025-09-16
add terraform version for v1.6.0, v1.5.1, and v1.5.0

FrankYang0529:


Vicente-Cheng (approved):
thanks for adding these!

PR #894, opened 2025-10-08
docs: expand info about FS freeze for VM Backup/Snapshot

WebberHuang1118:
Expand info about FS freeze for VM Backup/Snapshot

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9241

#### Additional Information:
@jillian-maroket Because filesystem freeze isn’t specific to VM backups and can also be performed for VM snapshots, I separated the topic into its own section instead of keeping it under the VM backup content.

Additionally, we should emphasize that the filesystem freeze functionality depends on the guest VM’s configuration and isn’t controlled by Harvester. Some users may not be aware of this, so we need to state it explicitly.

Thanks


Vicente-Cheng (approved):
LGTM, thanks for the enhancement!

docs/vm/backup-restore.md
```diff
@@ -372,6 +305,104 @@ Volumes consume extra disk space in the cluster whenever you create a new virtua
...
+  kubectl exec -it $POD -n default -c compute -- bash
+  ```
+
+1. Attempt to freeze the filesystem using the virt-freezer application, which is available in the `compute` container:
+
+  ```bash
+  virt-freezer --freeze --namespace <VM namespace> --name <VM name>
+  ```
+
+1. Verify the result of the freeze operation.
```
ihcsim: I think it's worth repeating the reminder to unfreeze the file system here, even though it's already included in the troubleshooting section, just in case the reader didn't read that far.

WebberHuang1118: Hi @ihcsim 

We highlighted the importance of thawing the filesystem, as suggested by @jillian-maroket.
```
  :::info important

  Do not skip this step. In addition, you must thaw virtual machine filesystems before performing any other operations.

  :::

```

ihcsim (approved):

PR #921, opened 2025-11-18
Add a note for the disk removal rejection

martindekov:
Adding a note in the beginning of the Remove disks section to mention that in case the disk to be removed hosts the last replica, deletion of thedisk will be rejected

#### Problem:
In 1.7 we no longer allow deletion of disks containing last healthy replicas and it can be confusing for the user.

#### Solution:
Add a note in the beginning of the `Remove disks` section to elaborate why the rejection might happen.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/3344#issuecomment-3540661397

#### Test plan:
N/A

N/A

### DIFF
--- a/docs/host/host.md
+++ b/docs/host/host.md
@@ -268,13 +268,13 @@ Before removing a disk, you must first evict Longhorn replicas on the disk.
 
 :::note
 
-The replica data would be rebuilt to another disk automatically to keep the high availability. There is a known issue on v2 where single replicated volume will be attempted
+The replica data would be rebuilt to another disk automatically to keep the high availability.
 
 :::
 
 :::note
 
-If disc contains last healthy replica of volume or backing image, disk removal will be rejected. In case a single replica of those objects was configured, move the replica to another disk or remove the objects  altogether to acknowledge that you don't need them. In case there was fault in the replication, fix the faulty replicas on the other disks before proceeding with the deletion of the current one. There is a known issue with multiple [Longhorn V2](https://github.com/longhorn/longhorn/issues/12189) disks where removal of the disk containing single replica blank volume will be rejected if it was once attached to another disk.
+If disc contains last healthy replica of volume or backing image, disk removal will be rejected. In case a single replica of those objects was configured, move the replica to another disk or remove the objects altogether to acknowledge that you don't need them. In case there was fault in the replication, fix the faulty replicas on the other disks before proceeding with the deletion of the current one. For a Longhorn v2 volume that has never been attached, if its last replica is deleted, Longhorn incorrectly triggers a rebuild on another disk. This rebuilt replica is then marked as healthy, causing the disk to become undeletable even though it contains no valid data tracked in [Issue #12189](https://github.com/longhorn/longhorn/issues/12189). In order to delete the disk, remove the volume first.
 
 :::
 
--- a/docs/host/host.md
+++ b/docs/host/host.md
@@ -268,13 +268,14 @@ Before removing a disk, you must first evict Longhorn replicas on the disk.
 
 :::note
 
-The replica data would be rebuilt to another disk automatically to keep the high availability.
+The replica data would be rebuilt to another disk automatically to keep the high availability. There is a known issue on v2 where single replicated volume will be attempted
 
 :::
 
 :::note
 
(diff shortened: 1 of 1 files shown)


akashraj4261 (commented):
Minor comment

w13915984028 (dismissed):
LGTM regarding the technical description, thanks.

BTW: for the wording, please wait for the doc experts' approval.

dariavladykina (commented):
Hi, please see a linguistic suggestion for this paragraph. Thanks!

docs/host/host.md
```diff
@@ -272,6 +272,12 @@ The replica data would be rebuilt to another disk automatically to keep the high
 
 :::
 
+:::note
+
+If the disc contains last healthy replica of volume or backing image, disk removal is rejected. In case a single replica of those objects is configured, move the replica to another disk or remove the objects altogether to acknowledge that you don't need them. In case there is a replication fault, fix the faulty replicas on the other disks before proceeding with the deletion of the current one.
```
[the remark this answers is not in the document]
martindekov: Sounds good, fixed

w13915984028 (dismissed):
LGTM, thanks.

WebberHuang1118 (commented):
We should mention the known issue for LH v2 engine https://github.com/harvester/harvester/issues/3344#issuecomment-3540661397

docs/host/host.md
```diff
@@ -268,7 +268,14 @@ Before removing a disk, you must first evict Longhorn replicas on the disk.
...
 
-The replica data would be rebuilt to another disk automatically to keep the high availability.
+The replica data would be rebuilt to another disk automatically to keep the high availability. There is a known issue on v2 where single replicated volume will be attempted
+
+:::
+
+:::note
+
+If disc contains last healthy replica of volume or backing image, disk removal will be rejected. In case a single replica of those objects was configured, move the replica to another disk or remove the objects  altogether to acknowledge that you don't need them. In case there was fault in the replication, fix the faulty replicas on the other disks before proceeding with the deletion of the current one. There is a known
+issue with [Longhorn V2](https://github.com/longhorn/longhorn/issues/12189) where removal of the disk containing volume with single replica will be rejected as the controller is trying to rebuild the replica on the disk.
```
WebberHuang1118: If I understand correctly, the context here doesn’t align with the known issue found by @albinsun in [this discussion](https://github.com/harvester/harvester/issues/3344#issuecomment-3473703045).

albinsun: The found issue is about on LHv2, a blank volume (Detached, never used) experienceing reattach to another disk will be considered as the **last healthy replica** in this enhancement. This behavior is not expected, track by longhorn/longhorn/issues/12189. 

For example
1. A **single replicated blank volume** on **multi-disks** using **LHv2** data engine 
1. When delete the (current/original) disk contains the last replica, LH will attach this volume to the (new/another) LHv2 disk and execute **rebuilding**.
1. Previous unexpected behavior makes the blank volume satisfies **last healthy replica** condition on harvester/harvester/issues/3344, so the user can NOT delete the disk.

martindekov: Thanks for the input @albinsun, reworded as per ^

albinsun (dismissed):
LGTM

@WebberHuang1118 it should be ok from LH perspective, do you think it's ok for harvester users?

docs/host/host.md
```diff
@@ -268,7 +268,13 @@ Before removing a disk, you must first evict Longhorn replicas on the disk.
 
 :::note
 
-The replica data would be rebuilt to another disk automatically to keep the high availability.
+The replica data would be rebuilt to another disk automatically to keep the high availability. There is a known issue on v2 where single replicated volume will be attempted
```
w13915984028: is this expanded note dedicated for head branch, or is also fit for existing ones like v1.6, v1.5? when true, please  copy to them

w13915984028 (dismissed):
LGTM, thanks.

WebberHuang1118 (dismissed):
LGTM — thanks for the explanation.

w13915984028 (dismissed):
LGTM, thanks.

albinsun (dismissed):
LGTM, thank you Martin.

Vicente-Cheng (dismissed):
The content looks good to me, thanks!

dariavladykina (approved):

PR #952, opened 2026-01-14
enhancement: drop BIOS support for new installs

tserong:
#### Problem:
We need to eventually remove support for legacy BIOS systems. Ensuring new installs can only be performed on UEFI systems is part of that journey.

#### Solution:
Explain that the installer ISO will only boot on UEFI systems, and that new installs are only possible on UEFI systems. Keep the mention that existing BIOS systems will continue to operate correctly.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9854
https://github.com/harvester/harvester/issues/9855

#### Test plan:
N/A

Right now (2026-01-14), the change to drop BIOS support from the installer ISO has been merged for v1.8.0, but it's still possible to install on BIOS systems via PXE boot. Once the latter is also gone, I'll open another docs PR with a warning on the PXE Boot Installation page indicating what sort of errors will occur when trying to install on BIOS systems via PXE, and also to drop the mention of the `install.force_mbr` setting which will then be redundant.


jillian-maroket (approved):
Thanks!

PR #966, opened 2026-01-28
Add a known issue to v170

w13915984028:
#### Solution:

Add this known issue to the document

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9751

### DIFF
--- a/docs/upgrade/v1-6-x-to-v1-7-x.md
+++ b/docs/upgrade/v1-6-x-to-v1-7-x.md
@@ -232,3 +232,53 @@ This workaround is only necessary when upgrading to v1.7.0. In v1.7.1 and later 
 :::
 
 Related issues: [#9815](https://github.com/harvester/harvester/issues/9815) and [#9802](https://github.com/harvester/harvester/issues/9802)
+
+### 4. After upgrade from to v1.7.x, Running VM show message "Restart action is required ..."
+
+After upgrade, the Harvester UI might show message like `"Restart action is required ..."`, Kubevirt adds a field to the vm definition implicitly during the upgrade.
+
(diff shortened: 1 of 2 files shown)


docs/upgrade/v1-6-x-to-v1-7-x.md
```diff
@@ -232,3 +232,53 @@ This workaround is only necessary when upgrading to v1.7.0. In v1.7.1 and later
 :::
 
 Related issues: [#9815](https://github.com/harvester/harvester/issues/9815) and [#9802](https://github.com/harvester/harvester/issues/9802)
+
+### 4. After upgrade from to v1.7.x, Running VM show message "Restart action is required ..."
+
+After upgrade, the Harvester UI might show message like `"Restart action is required ..."`, Kubevirt adds a field to the vm definition implicitly during the upgrade.
```
jillian-maroket:
````suggestion
### 4. Restart Required for Running Virtual Machines

After the upgrade, the Harvester UI may show the message `"Restart action is required ..."` for running virtual machines. This occurs because KubeVirt adds the following field to the virtual machine definition during the upgrade.

```yaml
spec:
  template:
    spec:
      domain:
          firmware:
            uuid: <VALUE>
```
````

jillian-maroket (commented):
Review done

Vicente-Cheng (approved):
contents LGTM, 
just once question, should we also need to update to https://docs.harvesterhci.io/v1.8/upgrade/v1-7-x-to-v1-7-y?

w13915984028:
From v17*, kubevirt appends those fields automatically when VM is created, and will not cause change in the upgrade path to newer version.

PR #967, opened 2026-02-02
Use stage name "initramfs" when setting up multipathd

tserong:
#### Problem:
The `default` stage isn't actually a thing.

#### Solution:
Replace with `initramfs`. Also fixed a small typo.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7469

#### Test plan:
N/A


jillian-maroket (approved):
Thanks!

PR #976, opened 2026-02-17
update kubeovn-vpc doc to document dns_server settings

ibrokethecloud:
#### Problem:

PR details how to add dns_server at the subnet level to allow VM workloads to leverage the dns server without having to explicitly specify one via cloudinit.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9687


rrajendran17 (dismissed):

starbops (commented):
I am a bit hesitant about this, but since the proposed addition note is right after the "Edit as YAML" step, would it be handy if we could also provide a YAML snippet example for the `dhcpV4Options` field? We can leave this as is for sure, but switching to YAML mode for editing the `ensureDHCP` field and jumping back to the original mode for the `DHCPV4Options` field feels bad in terms of UX.

WDYT?

starbops (approved):
LGTM, thank you.

akashraj4261 (approved):

PR #987, opened 2026-03-11
Add ssh password authentication config explanation

martindekov:
Adding reference and explanation for the new feature of the installer which enables you to enable or
disable ssh authentication through password for 1.8+

#### Problem:
Misleading reference for sshd after latest feature being added

#### Solution:
Add all relevant fields for 1.8+ version

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8548

#### Test plan:
https://69b137c84d43a8bc6a162643--harvester-preview.netlify.app/v1.8/install/harvester-configuration/#ossshd


ihcsim (dismissed):

docs/install/harvester-configuration.md
```diff
@@ -501,18 +501,19 @@ os:
...
-Subsystem used to configure the OpenSSH Daemon (sshd). Harvester currently only supports `sftp`.
+Settings used to configure the OpenSSH Daemon (sshd). You can enable SFTP file transfer using the `sftp` setting. Post-installation SSH password authentication can be disabled using the `disable_password_auth` setting.
 
 #### Example
 
 ```yaml
 os:
   sshd:
     sftp: true  #  The SFTP subsystem is enabled.
+    disable_password_auth: true  #  SSH password authentication is disabled.
```
w13915984028: please also mention this takes effects only when a ssh key is existing

and update https://docs.harvesterhci.io/v1.8/install/index step 14

BTW, if user uses a configure file (e.g. pxe mode), where key is not set but password_auth is disabled, what happens? thanks.

ihcsim: Yeah, I sorta wondered why we needed that for ISO mode when I first reviewed it, but thought it was also ok.

If user truly locked themselves out, they should have some out-of-band remote access to their server over serial port.

martindekov: I see, thanks for clarifying this Jian, didn't know you could still connect to the VM via console if ssh is disabled. Will go with option 2 to document the limitation, that way people can truly disable all ssh access and in case someone desires otherwise we can implement it.

ihcsim (dismissed):

bk201 (dismissed):
thank you!

akashraj4261 (dismissed):

w13915984028 (approved):
LGTM, thanks.

PR #990, opened 2026-03-20
Add docs for rwx-network

brandboat:
docs for https://github.com/harvester/harvester/issues/7218 and https://github.com/harvester/harvester/issues/6910

### DIFF
--- a/docs/advanced/rwxnetwork.md
+++ b/docs/advanced/rwxnetwork.md
@@ -1,0 +1,357 @@
+---
+sidebar_position: 14
+sidebar_label: RWX Network
+title: "RWX Network"
+---
+
+<head>
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.8/advanced/rwxnetwork"/>
+</head>
+
+_Available as of v1.8.0_
+
+Harvester uses Longhorn to provide ReadWriteMany (RWX) volumes for workloads that require shared filesystem access. By default, RWX traffic uses the Kubernetes cluster network. You can configure the [`rwx-network`](./settings.md#rwx-network) setting to either share it with the existing [storage network](./storagenetwork.md), or isolate RWX traffic on a dedicated network for better network bandwidth, performance, and security.
+
+:::note
+
+- Avoid configuring Longhorn settings directly, as this can result in unexpected or unwanted system behavior.
+
+:::
+
+## Prerequisites
+
+Before you begin configuring the RWX network, ensure that the following requirements are met:
+
+- The [cluster network](../networking/clusternetwork.md) and [VLAN network](../networking/harvester-network.md#vlan-network) are configured correctly. Ensure that both networks cover all nodes and are accessible.
+
+- If using **share mode** (`share-storage-network: true`), the [storage network](./storagenetwork.md) must be configured first.
+
+  When share mode is enabled, the `longhorn-csi-plugin` and `share-manager` pods also use the storage network and consume IPs from its range. Ensure the storage network IP range accounts for these additional pods. For details, see the [storage network prerequisites](./storagenetwork.md#prerequisites).
+
+- If using **dedicated mode** (`share-storage-network: false` and `network` set), the IP range of the RWX network has the following characteristics:
+
+  - Uses the IPv4 CIDR format
+
+  - Does not conflict or overlap with Kubernetes cluster networks, storage network, or VM migration network
+
+    The following addresses are reserved: `10.42.0.0/16`, `10.43.0.0/16`, `10.52.0.0/16`, and `10.53.0.0/16`.
+
+  - Covers the requirements of the cluster
+
+    The following pods use the RWX network and each require one IP address:
+
+    - `longhorn-csi-plugin` pods: One pod per node (DaemonSet), always present.
+    - `share-manager` pods: One pod per active RWX volume, created on demand.
+
+    The required number of IP addresses is calculated using the following formula: `Required number of IPs = Number of nodes + 32`, where 32 is an arbitrary buffer to accommodate current and future `share-manager` pods.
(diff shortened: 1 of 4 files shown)


w13915984028 (dismissed):
LGTM, thanks for the detailed document.

akashraj4261 (dismissed):

docs/advanced/rwxnetwork.md
```diff
@@ -0,0 +1,367 @@
...
+
+  You can check if the `ippools.whereabouts.cni.cncf.io` CRD exists in the cluster using the command `kubectl get crd ippools.whereabouts.cni.cncf.io`.
+
+  If an empty string is returned, add the CRDs in [this directory](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds) using the following commands:
+
+  ```
+  kubectl apply -f https://raw.githubusercontent.com/harvester/harvester/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds/whereabouts.cni.cncf.io_ippools.yaml
+
+  kubectl apply -f https://raw.githubusercontent.com/harvester/harvester/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
+  ```
```
WebberHuang1118: Do we need to mention this part specifically? Whereabouts should already be deployed.

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks.

docs/advanced/rwxnetwork.md
brandboat: Sure

brandboat: Removed in the latest commit, do u want to take another look?

docs/advanced/rwxnetwork.md
```diff
@@ -0,0 +1,355 @@
...
+  
+  ```bash
+  kubectl get volumes.longhorn.io -n longhorn-system -o json | jq '.items[] | select(.spec.accessMode == "rwx" and .spec.migratable == false) | {name: .metadata.name, state: .status.state}'
+  ```
+
+## `rwx-network` Setting
+
+The [`rwx-network`](./settings.md#rwx-network) setting allows you to configure the network used to isolate RWX volume traffic when segregation is required.
+
+You can [enable](#enable-the-rwx-network) and [disable](#disable-the-rwx-network) the RWX network using either the UI or the CLI. The setting value is a JSON object with the following structure:
```
ihcsim:
```suggestion
You can [enable](#enable-the-rwx-network) and [disable](#disable-the-rwx-network) the RWX network using either the UI or the CLI. The setting value is represented by a JSON object with the following structure:
```

docs/advanced/rwxnetwork.md
```diff
@@ -0,0 +1,355 @@
...
+{
+  "share-storage-network": true|false,
+  "network": {
+    "vlan": <vlan-id>,
+    "clusterNetwork": "<cluster-network-name>",
+    "range": "<cidr-range>",
+    "exclude": ["<excluded-ip-or-range>"]
+  }
+}
+```
```
ihcsim, line 82:
````suggestion
```

To configure a _shared_ RWX network, set `share-store-network` to `true`. Setting this field to `false` with a non-empty `network` field enables a _dedicated_ RWX network.
```

File: docs/advanced/rwxnetwork.md [hunk 4]
```diff
@@ -0,0 +1,355 @@
...
+
+Before you begin configuring the RWX network, ensure that the following requirements are met:
+
+- The [cluster network](../networking/clusternetwork.md) and [VLAN network](../networking/harvester-network.md#vlan-network) are configured correctly. Ensure that both networks cover all nodes and are accessible.
+
+- If using **share mode** (`share-storage-network: true`), the [storage network](./storagenetwork.md) must be configured first.
+
+  When share mode is enabled, the `longhorn-csi-plugin` and `share-manager` pods also use the storage network and consume IPs from its range. Ensure the storage network IP range accounts for these additional pods. For details, see the [storage network prerequisites](./storagenetwork.md#prerequisites).
+
+- If using **dedicated mode** (`share-storage-network: false` and `network` set), the IP range of the RWX network has the following characteristics:
```
ihcsim, line 31: Are there any differences in the IP range requirements for the shared and dedicated modes? If they truly are the same, can we just add a hyperlink from here to the IP range prerequisite on the storage network page?

I find this information tedious and repetitive. It will be nice to keep them in one place.

For example, on this page, say something like:

```
The IP range chosen for the RWX network must satisfy the [IP range requirements defined for the storage network](./storagenetwork.md/#ip-range-requirements)."
```

On the storage network page, we organize the prequisites slightly differently:


```md
# Prequisites

<existing prerequisites>

## IP Range Requirements

* The IP range of the storage network has the following characteristics:
    * Uses the IPv4 CIDR format
    * Does not conflict or overlap with Kubernetes cluster networks
    * etc.
```

brandboat: I totally get that the repetition is annoying. But I think keeping them separate is actually a necessary evil for now. Even though the basic rules look the same, the IP allocation logic and the Longhorn/component requirements for the RWX network are actually different from the Storage network.

Besides that https://docs.harvesterhci.io/v1.7/advanced/vm-migration-network/#prerequisites the vm migration network docs also encounter the same issue, perhaps we can find a way to standardize these 'common' prerequisites without losing the specific technical details. But this could be done in a separate pr IMO.

ihcsim (approved):

WebberHuang1118 (approved):
LGTM, thanks for the documentation.

PR #1002, opened 2026-03-30
docs: clarify server_url on cluster creation

fluktuid:
the creating node may not have a server_url set
this is due to harvester-installer config validator the config-valudator needs to do this thus harvester checks the value

#### Problem:
when creating a harvester cluster via config (e.g. via ipxe) `server_url` must not be present (or with value)
the current docs are a bit unclear regarding this.

#### Solution:
extend documentation

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10325

#### Test plan:
not applicable

see https://github.com/harvester/harvester-installer/blob/8317a66fd50b7954f8d1a87a957abbb11994bfbc/pkg/console/validator.go\#L422
see https://github.com/harvester/harvester-installer/blob/8317a66fd50b7954f8d1a87a957abbb11994bfbc/pkg/config/cos.go\#L377

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -21,7 +21,7 @@ Harvester configuration file can be provided during manual or automatic installa
 
 ```yaml
 scheme_version: 1
-server_url: https://cluster-VIP:443
+server_url: https://cluster-VIP:443 # empty or absent in create mode
 token: TOKEN_VALUE
 os:
   ssh_authorized_keys:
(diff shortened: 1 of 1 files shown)


akashraj4261 (approved):

brandboat:
gentle ping @bk201, could you please help review this PR when you're available? thanks!

tserong (approved):

tserong:
@fluktuid can you please apply the same change to `versioned_docs/version-v1.8/install/harvester-configuration.md`? That way it'll appear in the docs for v1.8 as well, which is the default view when you visit https://docs.harvesterhci.io/

jillian-maroket:
@tserong Does the v1.7 doc require any changes? Not sure which versions are affected by this.

tserong:
@jillian-maroket actually, yeah, this applies to all versions

jillian-maroket:
I'll merge this now and create another PR for backporting.

tserong:
Thanks @jillian-maroket!

PR #1014, opened 2026-04-16
docs: update workaround note

DavideRutigliano:
#### Problem:

Note in [workaround#3](https://docs.harvesterhci.io/v1.7/upgrade/v1-6-x-to-v1-7-x#3-persistent-names-of-certain-network-interfaces-may-change-during-upgrade) for Harvester upgrade from 1.6.x to 1.7.x does not cover the case with bonded interfaces, leading to ambiguities and possible issues after upgrade.

#### Solution:

Add clarification on interface renaming workaround, specifying cases where workaround is required vs automated.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10397

### DIFF
--- a/docs/upgrade/v1-6-x-to-v1-7-x.md
+++ b/docs/upgrade/v1-6-x-to-v1-7-x.md
@@ -227,11 +227,17 @@ To resolve this issue, apply kernel arguments that restore the original names of
 
 :::note
 
-This workaround is only necessary when upgrading to v1.7.0. In v1.7.1 and later versions, these `ifname=` arguments are automatically added to prevent network disruptions during driver updates.
+This workaround is only necessary when upgrading to v1.7.0 or if using bonded interfaces. In v1.7.1 and later versions, these `ifname=` arguments are automatically added to prevent network disruptions during driver updates.
+
+| Version | Interface | Workaround |
+| --- | --- | --- |
+| v1.7.0 | Any | Required |
+| v1.7.1+ | Standalone | Automated (No action) |
+| v1.7.1+ | Bonded | Required |
 
 :::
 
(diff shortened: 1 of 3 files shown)


docs/upgrade/v1-6-x-to-v1-7-x.md
```diff
@@ -227,11 +227,17 @@ To resolve this issue, apply kernel arguments that restore the original names of
 
 :::note
 
-This workaround is only necessary when upgrading to v1.7.0. In v1.7.1 and later versions, these `ifname=` arguments are automatically added to prevent network disruptions during driver updates.
+This workaround is only necessary when upgrading to v1.7.0 or if using bonded interfaces. In v1.7.1 and later versions, these `ifname=` arguments are automatically added to prevent network disruptions during driver updates.
+
+| Version | Interface | Workaround |
+| --- | --- | --- |
+| v1.7.0 | Any | Required |
+| v1.7.1+ | Standard | Automated (No action) |
```
tserong: I'd suggest "Standalone" instead of "Standard" here (at least, that's how I think of an interface that's not bonded - I'm not sure if there's other or better terms for this)

tserong (commented):
Thanks for this! Just one small suggestion.

We'll also need to update `versioned_docs/version-v1.7/upgrade/v1-6-x-to-v1-7-x.md` to include this change.

akashraj4261 (approved):

tserong:
This same change still needs to be applied to the file versioned_docs/version-v1.7/upgrade/v1-6-x-to-v1-7-x.md or it'll only appear in the dev docs (not the v1.7 docs)

tserong (approved):

PR #1021, opened 2026-04-23
enhancement: drop BIOS support for new installs (PXE)

tserong:
#### Problem:
We need to eventually remove support for legacy BIOS systems. Ensuring new installs can only be performed on UEFI systems is part of that journey.

#### Solution:
- Explain that PXE install is no longer possible on legacy BIOS systems
- Drop mention of the `install.force_mbr` config option
- A minor wording change to unify the notes with the ISO install

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9854
https://github.com/harvester/harvester/issues/9855

#### Test plan:
N/A

This is a followup to https://github.com/harvester/docs/pull/952

### DIFF
--- a/docs/install/harvester-configuration.md
+++ b/docs/install/harvester-configuration.md
@@ -90,7 +90,6 @@ install:
   cluster_pod_cidr: 10.52.0.0/16
   cluster_service_cidr: 10.53.0.0/16
   cluster_dns: 10.53.0.10
-  force_mbr: false
   addons:
     harvester_vm_import_controller:
       enabled: false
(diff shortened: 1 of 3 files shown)


ihcsim:
There are two more places that have references to "BIOS". Do we want to remove them too?

```sh
$ grep -rniI --exclude-dir=api bios docs/install
docs/install/pxe-boot-install.md:430:To perform PXE boot over tagged VLAN network, the Harvester hosts must be configured with the following BIOS/UEFI settings:
docs/install/external-disk-support.md:32:Necessary changes to the BIOS or firmware will depend on the hardware that you use.
```

tserong:
@ihcsim Yeah... Uhm... Those are talking about going into the system BIOS setup utility to tweak the host config in some way. I don't know if there's a better way to word that. I mean, I think of the BIOS setup utility as being the thing that you use to switch between legacy BIOS boot mode and UEFI boot mode. But the fact that the term BIOS gets used here could arguably be a bit confusing. I just don't know if there's a better term to use...?

akashraj4261 (approved):

ihcsim (approved):

wheatdog (approved):
Thanks!

Vicente-Cheng (approved):
thanks!

PR #1026, opened 2026-04-27
Add troubleshooting for issue 10188 guest cluster stucks provisioning

w13915984028:
#### Problem:

#10188 is a known issue on v180.

#### Solution:

Add details and workaround to document.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10188


wheatdog (approved):
Thanks!

PR #1041, opened 2026-05-08
doc: update NIC renaming workaround for v1.7.2

tserong:
#### Problem:
The NIC renaming workaround is no longer required in v1.7.2

#### Solution:
Update the v1.6.x->v1.7.x upgrade page

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10397

#### Test plan:
N/A

N/A


jillian-maroket (approved):
Thanks!

PR #1046, opened 2026-05-20
docs: update export restriction from volume to image in volume security documentation

Yu-Jack:
https://github.com/harvester/harvester/issues/10596


WebberHuang1118 (approved):
LGTM, thanks.

PR #1051, opened 2026-05-28
Clarify that RWX filesystem volumes require migratable=false with dataEngine=v2

glovecchi0:
#### Problem:
When using dataEngine=v2, setting migratable=true creates a migratable RWX block volume intended for VM live migration workflows instead of a standard RWX filesystem volume.

As a result, the standard Longhorn RWX sharing mechanism based on share-manager and NFS is not enabled, causing RWX filesystem volumes to not work as expected.

To use standard RWX filesystem volumes with dataEngine=v2, migratable must remain unset or be explicitly set to false.

#### Solution:
Document that `migratable=true` must not be used with `dataEngine=v2` for standard RWX filesystem volumes, and clarify that `migratable=false` (default) is required to enable the Longhorn share-manager and NFS-based RWX sharing layer.


glovecchi0:
CC: @Vicente-Cheng , @Martin-Weiss

Martin-Weiss:
I guess the "migratable=false" is required in general for all "RWX file" regardless of the engine?

glovecchi0:
@Martin-Weiss, let me open a new PR, as I've done a mistake with the signing.

PR #1065, opened 2026-06-18
Fix a dead link

w13915984028:
#### Problem:

When click the link [the IP pool selection policy](https://docs.harvesterhci.io/networking/ippool.md/#selection-policy)  on page https://docs.harvesterhci.io/v1.8/networking/loadbalancer#how-to-create

image

it can't be opened, as it is missing the version info

#### Solution:

Fix the dead link


martindekov (approved):
LGTM

PR #1070, opened 2026-06-30
docs(upgrade): add known issue for SUC prepare plan retry failure

starbops:
#### Problem:

SUC (system-upgrade-controller) might stop working on newly created Plans out of a blue. The root cause is still unknown, as it cannot be reproduced steadily. Harvester Upgrade relies on SUC to perform node-level tasks, one of the most well-known ones is the `prepare` Plan, which preloads container images to each node.

If such a Plan does not get reconciled by SUC, the entire Harvester Upgrade procedure stalls.

#### Solution:

By manually restarting the SUC deployment, the upgrade should resume immediately.

#### Related Issue(s):

harvester/harvester#9880


wheatdog (approved):
Thanks!

bk201 (approved):

jillian-maroket (approved):

PR #1081, opened 2026-07-13
Add note for host network config behaviour

rrajendran17:
#### Problem:
The issue occurs when the host is attached multiple times to the same Layer 2 network. In the VLAN 1 case, mgmt-br already represents the native/untagged VLAN 1 network, and creating mgmt-br.1 adds a second host interface to the same broadcast domain. DHCP DISCOVER packets are transmitted successfully, but the DHCP OFFER is associated with the existing interface (mgmt-br) rather than the VLAN subinterface (mgmt-br.1), causing the DHCP client to never complete lease acquisition.

More generally, a node should not have multiple host interfaces connected to the same VLAN/subnet

#### Solution:
document the behaviour so users know what to configure

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10802

### DIFF
--- a/docs/networking/hostnetworkconfig.md
+++ b/docs/networking/hostnetworkconfig.md
@@ -319,7 +319,7 @@ Delete the config using the following command:
 
 :::caution
 
-If a node has multiple host interfaces, do not connect them to the same VLAN or subnet.
+If a node has multiple host interfaces, do not attach them to the same VLAN or subnet. This restriction applies to interfaces configured manually on the host and those configured through `HostNetwork` configurations.
 
 VLAN interfaces under a cluster network are derived from the same bridge and therefore use the same MAC address. This can cause the following issue:
 - **DHCP response misrouting**: When multiple host interfaces from the same cluster network reside on the same VLAN, DHCP responses may be associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This occurs because the DHCP `OFFER` packets can be delivered to `mgmt-br` instead of `mgmt-br.1`.
--- a/docs/networking/hostnetworkconfig.md
+++ b/docs/networking/hostnetworkconfig.md
@@ -317,6 +317,22 @@ Delete the config using the following command:
 
 `kubectl delete hostnetworkconfig cn1-vlan2012-dhcp`.
 
+:::caution
+
+If a node has multiple host interfaces, do not connect them to the same VLAN or subnet.
+
+VLAN interfaces under a cluster network are derived from the same bridge and therefore use the same MAC address. This can cause the following issue:
+- **DHCP response misrouting**: When multiple host interfaces from the same cluster network reside on the same VLAN, DHCP responses may be associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This occurs because the DHCP `OFFER` packets can be delivered to `mgmt-br` instead of `mgmt-br.1`.
+
+Configuring multiple interfaces in the same VLAN and subnet can result in ambiguous routing behavior.
+- **Ambiguous kernel routing**: Linux typically installs a single connected route for a subnet. When multiple interfaces are attached to the same VLAN and subnet, the kernel may associate the subnet route with only one of those interfaces, causing traffic to be forwarded through an unintended path.
+
+For example, if a host network on cluster network `cn1` uses VLAN `2017` and receives subnet `192.168.0.0/24`, while another host network on cluster network `cn2` is also configured with VLAN `2017` and receives the same subnet 192.168.0.0/24, the kernel will install the connected route for 192.168.0.0/24 via only one interface (the latest configured). As a result, traffic destined for that subnet may be routed through the wrong interface, leading to unexpected issues.
+
+This restriction applies to interfaces configured both manually on the host and through `HostNetwork` configurations.
+
+:::
+
 ### Limitations
 
 - Only tagged VLANs with a VLAN ID of 1–4094 are supported.
--- a/docs/networking/hostnetworkconfig.md
+++ b/docs/networking/hostnetworkconfig.md
@@ -327,12 +327,11 @@ VLAN interfaces used in a cluster network are derived from the same underlying b
 
 When multiple host interfaces from the same cluster network reside on the same VLAN, DHCP responses may be incorrectly associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an active IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This failure occurs because the DHCP `OFFER` packets are delivered to `mgmt-br` instead of `mgmt-br.1`.
 
-Configuring multiple interfaces in the same VLAN and subnet can result in ambiguous routing behavior.
-- **Ambiguous kernel routing**: Linux typically installs a single connected route for a subnet. When multiple interfaces are attached to the same VLAN and subnet, the kernel may associate the subnet route with only one of those interfaces, causing traffic to be forwarded through an unintended path.
+**Ambiguous Routing Behavior**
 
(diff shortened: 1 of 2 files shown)


versioned_docs/version-v1.8/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,16 @@ Delete the config using the following command:
 
 `kubectl delete hostnetworkconfig cn1-vlan2012-dhcp`.
 
+:::note
+
+A node must not have multiple host interfaces connected to the same VLAN/subnet. Within the same cluster network, VLAN interfaces are derived from the same bridge and therefore use the same MAC address. When multiple host interfaces are attached to the same VLAN, DHCP responses may be associated with the existing interface instead of the newly created one. For example, if mgmt-br is already connected to the native VLAN and has an IP address, a DHCP client running on mgmt-br.1 may not successfully acquire a lease because DHCP OFFER packets can be delivered to mgmt-br rather than mgmt-br.1.
+
+In addition to DHCP issues, configuring multiple host interfaces in the same VLAN/subnet on a node can result in ambiguous routing behavior because the kernel installs routes for the subnet through a single interface, which may cause traffic to be routed through an unintended interface.
+
+The above applies to both host interfaces configured manually and through hostnetwork configurations.
```
jillian-maroket:
```suggestion
:::caution

If a node has multiple host interfaces, do not connect them to the same VLAN or subnet.

VLAN interfaces in a cluster network are derived from the same bridge and therefore use the same MAC address. This can cause the following issues:

- **DHCP response misrouting**: When multiple host interfaces reside on the same VLAN, DHCP responses may be associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This occurs because the DHCP `OFFER` packets can be delivered to `mgmt-br` instead of `mgmt-br.1`.

- **Ambiguous kernel routing**: Configuring multiple interfaces in the same VLAN or subnet can result in ambiguous routing behavior. The Linux kernel installs subnet routes through a single interface, which may force traffic through an unintended path.

This restriction applies to interfaces configured both manually on the host and through `HostNetwork` configurations.
```

jillian-maroket (commented):
Review done

docs/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,16 @@ Delete the config using the following command:
 
 `kubectl delete hostnetworkconfig cn1-vlan2012-dhcp`.
 
+:::note
+
+A node must not have multiple host interfaces connected to the same VLAN/subnet. Within the same cluster network, VLAN interfaces are derived from the same bridge and therefore use the same MAC address. When multiple host interfaces are attached to the same VLAN, DHCP responses may be associated with the existing interface instead of the newly created one. For example, if mgmt-br is already connected to the native VLAN and has an IP address, a DHCP client running on mgmt-br.1 may not successfully acquire a lease because DHCP OFFER packets can be delivered to mgmt-br rather than mgmt-br.1.
```
Yu-Jack: Could we do like - if user had created the host network config, then attached multiple same VLAN, we would show the warning in host network config resource?

rrajendran17: Thanks @Yu-Jack , webhook was my initial decision, but the problem is we need to get the existing interfaces on harvester host to compare the vlan with the host network config vlan id.
The webhook pod runs as deployment pod and host network is not enabled, so it will not be able to get the host interfaces. Even if we enable host network, since its a deployment pod it cannot get interfaces from all nodes.
So this will not be an option in webhook to validate.

Yeah other possibility is showing this as warning/error message in status from controller, but I thought these are network configuration errors so its better to educate the users through docs instead of handling those in controllers.
we will come across other misconfiguration from users in future also.

Yu-Jack: > Yeah other possibility is showing this as warning/error message in status from controller, but I thought these are network configuration errors so its better to educate the users through docs instead of handling those in controllers.
we will come across other misconfiguration from users in future also.

Actually it's a good idea to show the message in status via controller.

Here is my thought.

We provide two entries to users:

1. the message in status
2. this document

First option is for users that are encountering the issue.

Second option is for users that are going to  do this if they remember to check the document firstly.

However, since we have the deadline, it's not really necessary, just a suggestion.

docs/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,22 @@ Delete the config using the following command:
 
 `kubectl delete hostnetworkconfig cn1-vlan2012-dhcp`.
 
+:::caution
+
+If a node has multiple host interfaces, do not connect them to the same VLAN or subnet.
```
jillian-maroket:
```suggestion
If a node has multiple host interfaces, do not attach them to the same VLAN or subnet. This restriction applies to interfaces configured manually on the host and those configured through `HostNetwork` configurations.
```

docs/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,22 @@ Delete the config using the following command:
...
+
+VLAN interfaces under a cluster network are derived from the same bridge and therefore use the same MAC address. This can cause the following issue:
+- **DHCP response misrouting**: When multiple host interfaces from the same cluster network reside on the same VLAN, DHCP responses may be associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This occurs because the DHCP `OFFER` packets can be delivered to `mgmt-br` instead of `mgmt-br.1`.
```
jillian-maroket:
```suggestion
**DHCP Response Misrouting**

VLAN interfaces used in a cluster network are derived from the same underlying bridge and consequently share the same MAC address. This shared MAC address can cause DHCP response misrouting.

When multiple host interfaces from the same cluster network reside on the same VLAN, DHCP responses may be incorrectly associated with an existing interface instead of a newly created one. For example, if `mgmt-br` is connected to the native VLAN and has an active IP address, a DHCP client running on `mgmt-br.1` may fail to acquire a lease. This failure occurs because the DHCP `OFFER` packets are delivered to `mgmt-br` instead of `mgmt-br.1`.
```

docs/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,22 @@ Delete the config using the following command:
...
+
+Configuring multiple interfaces in the same VLAN and subnet can result in ambiguous routing behavior.
+- **Ambiguous kernel routing**: Linux typically installs a single connected route for a subnet. When multiple interfaces are attached to the same VLAN and subnet, the kernel may associate the subnet route with only one of those interfaces, causing traffic to be forwarded through an unintended path.
+
+For example, if a host network on cluster network `cn1` uses VLAN `2017` and receives subnet `192.168.0.0/24`, while another host network on cluster network `cn2` is also configured with VLAN `2017` and receives the same subnet 192.168.0.0/24, the kernel will install the connected route for 192.168.0.0/24 via only one interface (the latest configured). As a result, traffic destined for that subnet may be routed through the wrong interface, leading to unexpected issues.
```
jillian-maroket:
```suggestion
**Ambiguous Routing Behavior**

Additionally, configuring multiple interfaces in the same VLAN and subnet results in ambiguous routing. The Linux kernel typically installs a single connected route for a given subnet. If multiple interfaces are attached to the same VLAN and subnet, the kernel may associate the subnet route with only one of those interfaces, causing traffic to be forwarded through an unintended physical path.

Example: Host networks on cluster networks `cn1` and `cn2` are both attached to VLAN `2017` and assigned the same subnet (`192.168.0.0/24`). Because both interfaces are in the same subnet, the Linux kernel will install the connected route for `192.168.0.0/24` via only one of those interfaces (typically the one configured last). Traffic destined for that subnet may be routed through the wrong interface, leading to unexpected issues.
```

docs/networking/hostnetworkconfig.md
```diff
@@ -317,6 +317,22 @@ Delete the config using the following command:
...
+
+This restriction applies to interfaces configured both manually on the host and through `HostNetwork` configurations.
```
jillian-maroket: Please remove this line. I moved the sentence to paragraph 1.

jillian-maroket (commented):
The note is now very long because of the new text. However, I first need to evaluate the adjacent topics before creating a dedicated section for this. I still suggested using headings to break up the concepts even if we have to bold them for now (a crude solution).

jillian-maroket (approved):

ibrokethecloud (approved):

PR #1075, opened 2026-07-10
docs(lvm-local-storage): considerations for striped vs dm-thin

abonillabeeche:
## Summary

Adds two small pieces to the LVM Local Storage add-on documentation:

1. A brief **Considerations: `striped` vs `dm-thin`** section right after the StorageClass creation step, to help operators pick the right volume group type at the moment they are configuring it.
2. A small **Additional Notes** section at the bottom, covering:
   - Tuning knobs on the `dm-thin` pool (chunk zeroing, metadata size).
   - Recommendation to use `virtio-scsi` for VM disks backed by LVM CSI.
   - Coexistence with Longhorn v2 block-mode disks (LVM `global_filter`).

Same content is applied to both:
- `docs/advanced/addons/lvm-local-storage.md` (main)
- `versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md` (v1.8)

## Motivation

Both volume group types are already listed as supported, but the doc doesn't currently guide the reader on when to pick one over the other. In particular, the fact that `striped` snapshots reserve the origin's full capacity at creation time (versus `dm-thin` snapshots being effectively free at creation) is important for anyone planning to take regular snapshots on their VM workloads.

The Longhorn v2 coexistence note references [harvester/harvester#11098](https://github.com/harvester/harvester/issues/11098) for background.

## Notes

Documentation-only. No behavior changes.

PR #57, opened 2021-10-14
Add more OS troubleshooting tips

bk201:
- How to permanently edit kernel parameters
- How to change the default GRUB boot menu entry

**Related**
- https://github.com/harvester/harvester/issues/1225#issuecomment-942678991
- https://github.com/harvester/harvester/issues/1343#issuecomment-941380349

PR #1083, opened 2026-07-14
docs(lvm-local-storage): document chunk size + expand tuning section

abonillabeeche:
## Summary

Follow-up to [csi-driver-lvm PR #57](https://github.com/harvester/csi-driver-lvm/pull/57) — documents the new `chunkSize`, `poolMetadataSize`, and `zeroBlocks` StorageClass parameters and expands the existing "Tuning the `dm-thin` Pool" section (introduced in #1075) with:

- **NEW subsection: Chunk size** — the primary content of this PR. Explains why the default is `1M` (aligns with the RAID full-stripe width common on Harvester hardware, avoiding partial-stripe read-modify-write and, on parity RAID, write-hole exposure), when to override (match your RAID full stripe; `512K`/`128K` only for sub-stripe or non-RAID VGs, never >2M), the hard "cannot change after pool creation" constraint, and how to verify the effective chunk size of a live pool.
- **Chunk zeroing** — expanded to mention the `zeroBlocks` StorageClass parameter (in addition to the existing post-creation `lvchange --zero n` workflow).
- **Pool metadata size** — expanded to mention the `poolMetadataSize` StorageClass parameter (default `16G` covers 256 TB pools at 1M chunks) and adds the metadata sizing formula.
- **NEW: Full example StorageClass** — shows all three parameters set explicitly.

## Motivation

A field customer running everything from our runbook ended up with a 16 MiB chunk_size on their thin pool. LVM's default auto-select scales chunk_size with pool size to keep metadata bounded, which for multi-TB pools produces 8-16 MiB chunks — appropriate for metadata sizing but disastrous for random-IOPS workloads (up to 4096:1 allocation write amplification on 4K writes to virgin regions).

Customer's measured DI_RANDOM: **326 IOPS**. Our lab measurement on hardware in the same performance class with an explicit, stripe-aligned chunk size + `zero=n`: **15,000+ QD1 4K IOPS**. The gap is entirely explained by chunk size + zero-on-allocate.

## Test plan

- [x] Local Docusaurus render (`npm start`) — page renders, no broken links
- [x] Markdown syntax validated
- [ ] Docs CI (markdownlint + link check) — will run on this PR

## Base branch note

This PR includes the "Tuning the `dm-thin` Pool" section from PR #1075 (APPROVED + MERGEABLE) because it builds on that section. **Please merge #1075 first — this PR will show a clean diff (just the chunk-size additions) once it rebases against post-#1075 main.**

## Scope

Only edits `docs/advanced/addons/lvm-local-storage.md` (current-dev / v1.9 tree). Not backported to `versioned_docs/` because the `chunkSize` StorageClass parameter ships in a CSI driver version targeting Harvester v1.9. The `lvchange`/`lvextend` post-creation workflows work on older versions but were already covered by #1075.

## References

- [csi-driver-lvm PR #57](https://github.com/harvester/csi-driver-lvm/pull/57) — driver + chart changes this documents
- [LVM upstream `performance` policy](https://github.com/lvmteam/lvm2/blob/master/lib/config/defaults.h) — the 512 KiB general-purpose seed (Harvester defaults to `1M` for RAID full-stripe alignment)
- [Linux kernel dm-thin admin guide](https://docs.kernel.org/admin-guide/device-mapper/thin-provisioning.html) — "512 KiB general; 64 KiB snapshot-heavy"
- [Red Hat Gluster admin guide](https://docs.redhat.com/en/documentation/red_hat_gluster_storage/3.5/html/administration_guide/chap-configuring_red_hat_storage_for_enhancing_performance) — the 2 MiB cap "to reduce COW problems"

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -197,7 +197,7 @@ Backup creation is currently not supported. This limitation will be addressed in
 
 When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass, the driver creates an LVM thin pool named `<vgName>-thinpool` using `-l 90%FREE` (allocating 90% of the volume group's remaining free space). Consider tuning the following settings based on your workload demands:
 
-- **Chunk size**: The chunk size determines the smallest unit of physical space that a thin pool allocates in response to a write. A write to a previously-unallocated region always allocates a full chunk, so small random writes to virgin space against a large chunk size cause severe write amplification (a 4 KiB write against a 16 MiB chunk allocates 16 MiB of pool space). When you install the harvester-csi-driver-lvm add-on version 0.4.0 or later, the default StorageClass sets `chunkSize: "512K"`, which matches [LVM's upstream `performance` policy seed](https://github.com/lvmteam/lvm2/blob/master/lib/config/defaults.h) and the [Linux kernel dm-thin admin guide's general-purpose recommendation](https://docs.kernel.org/admin-guide/device-mapper/thin-provisioning.html), and comfortably supports thin pools up to 128 TB.
+- **Chunk size**: The chunk size determines the smallest unit of physical space that a thin pool allocates in response to a write. A write to a previously-unallocated region always allocates a full chunk, so small random writes to virgin space against a large chunk size cause severe write amplification (a 4 KiB write against a 16 MiB chunk allocates 16 MiB of pool space). When you install the harvester-csi-driver-lvm add-on version 0.4.0 or later, the default StorageClass sets `chunkSize: "1M"`. On the hardware-RAID-backed volume groups typical of Harvester nodes, `1M` matches the full-stripe width of common layouts (for example, four data disks at a 256 KiB strip yields a 1 MiB stripe), so each chunk allocation maps to whole stripes rather than partial ones. `1M` also keeps thin-pool metadata bounded and comfortably supports pools up to 256 TB. See the [Linux kernel dm-thin admin guide](https://docs.kernel.org/admin-guide/device-mapper/thin-provisioning.html) for background on chunk sizing.
 
   :::warning
 
@@ -205,13 +205,15 @@ When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass
 
   :::
 
+  Align the chunk size to your RAID full-stripe width — that is, (number of data disks) × (RAID strip size). A chunk smaller than the full stripe (for example, a `512K` chunk on a 1 MiB stripe) forces every first-touch allocation into a partial-stripe **read-modify-write**. On parity RAID (5/6), partial-stripe writes also widen the write-hole window, so an unclean shutdown without a protected controller cache (BBU/FBWC) can leave a stripe with inconsistent parity. Choosing a chunk size equal to — or an exact multiple of — the full stripe avoids both problems; `1M` is the safe default because it aligns with the most common power-of-two RAID geometries.
+
   Override the default with the `chunkSize` parameter on the StorageClass. Common values:
 
   | Value | When to use |
   |---|---|
-  | `128K` | Snapshot-heavy tenants (halves copy-on-write cost per touched region) |
-  | `512K` | **Default** — general-purpose VM workloads on pools up to 128 TB |
-  | `1M` | Pools larger than 128 TB, or bulk sequential-write workloads |
+  | `1M` | **Default** — matches the full-stripe width of common hardware-RAID layouts; general-purpose VM workloads |
+  | `512K` / `128K` | Only when the RAID full stripe is that size, or on non-RAID / single-disk volume groups where stripe alignment does not apply and minimizing snapshot copy-on-write cost is the priority |
+  | `2M` | Very large sequential-write pools whose RAID full stripe is 2 MiB |
 
   Never exceed `2M`. Larger chunks trigger disproportionate copy-on-write costs for snapshots, as documented in the [Red Hat Gluster admin guide](https://docs.redhat.com/en/documentation/red_hat_gluster_storage/3.5/html/administration_guide/chap-configuring_red_hat_storage_for_enhancing_performance). If you rely on LVM's built-in auto-selection instead of setting `chunkSize` explicitly, LVM chooses the chunk size based on the pool size to keep metadata bounded, which for multi-TB pools produces 8–16 MiB chunks — appropriate for metadata sizing but often not for random-write performance.
 
@@ -232,7 +234,7 @@ When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass
 
   You can fully reverse the change by running the command with `--zero y`.
 
-- **Pool metadata size**: When the thin pool is created, the driver sizes its metadata logical volume based on the `poolMetadataSize` StorageClass parameter (default `16G`, sufficient for pools up to 128 TB at the default 512 KiB chunk size). If you set `chunkSize` smaller than 512 KiB, the pool needs more metadata to address the same amount of data: metadata bytes ≈ 64 × (pool_size ÷ chunk_size). If you did not set the parameter, or if the pool grew significantly after creation, extend the metadata volume proactively to prevent the pool from running out of space and becoming unresponsive later.
+- **Pool metadata size**: When the thin pool is created, the driver sizes its metadata logical volume based on the `poolMetadataSize` StorageClass parameter (default `16G`, sufficient for pools up to 256 TB at the default 1 MiB chunk size). A larger chunk size needs less metadata to address the same capacity, and a smaller chunk size needs more: metadata bytes ≈ 64 × (pool_size ÷ chunk_size). If you did not set the parameter, or if the pool grew significantly after creation, extend the metadata volume proactively to prevent the pool from running out of space and becoming unresponsive later.
 
   ```
   sudo lvextend --poolmetadatasize +1G <vgName>/<vgName>-thinpool
@@ -249,7 +251,7 @@ When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass
   parameters:
     type: dm-thin
     vgName: vmvg
-    chunkSize: "512K"          # Applied on first-time pool creation only.
+    chunkSize: "1M"            # Applied on first-time pool creation only.
     poolMetadataSize: "16G"    # Applied on first-time pool creation only.
     zeroBlocks: "false"        # Applied on first-time pool creation only.
   reclaimPolicy: Delete
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -228,4 +228,4 @@ devices {
 }
 ```
 
-On Harvester's immutable OS, persist the change through an `/oem/*.yaml` cloud-config file so it survives reboots. For background, see [harvester/harvester#11098](https://github.com/harvester/harvester/issues/11098).
+Because Harvester's operating system is immutable, you must persist this change through an `/oem/*.yaml` cloud-config file to ensure it survives system reboots and upgrades.For more information, see issue [#11098](https://github.com/harvester/harvester/issues/11098).
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -207,7 +207,7 @@ Backup creation is currently not supported. This limitation will be addressed in
 
 When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass, the driver creates an LVM thin-pool named `<vgName>-thinpool` at `-l 90%FREE` of the volume group. Two settings are worth knowing about:
 
-- **Chunk zeroing.** By default, the pool writes zeros to each newly-allocated chunk before handing it to the writer. On single-tenant clusters this can be disabled to reduce write amplification on first-touch allocations:
+- **Chunk zeroing**: By default, the thin pool writes zeros to each newly allocated block chunk before exposing it to a write operation. On single-tenant clusters, you can disable chunk zeroing to significantly reduce write amplification during initial data allocations.
 
   ```
   sudo lvchange --zero n <vgName>/<vgName>-thinpool
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -213,9 +213,9 @@ When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass
   sudo lvchange --zero n <vgName>/<vgName>-thinpool
   ```
 
-  The change is fully reversible with `--zero y`.
+  You can fully reverse the change by running the command with `--zero y`.
 
-- **Pool metadata size.** LVM auto-sizes the thin-pool metadata LV at pool creation. For pools that will hold many snapshots or many thin volumes over time, extending the metadata proactively avoids running short later:
+- **Pool metadata size**: When the thin pool is created, LVM automatically sizes its metadata logical volume. Consider extending this volume proactively if you expect the pool to store a large number of snapshots or thin volumes over time. Doing so prevents the pool from running out of space and becoming unresponsive later.
 
   ```
   sudo lvextend --poolmetadatasize +1G <vgName>/<vgName>-thinpool
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -205,7 +205,7 @@ Backup creation is currently not supported. This limitation will be addressed in
 
 ### Tuning the `dm-thin` Pool
 
-When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass, the driver creates an LVM thin-pool named `<vgName>-thinpool` at `-l 90%FREE` of the volume group. Two settings are worth knowing about:
+When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass, the driver creates an LVM thin pool named `<vgName>-thinpool` using `-l 90%FREE` (allocating 90% of the volume group's remaining free space). Consider tuning the following settings based on your workload demands:
 
 - **Chunk zeroing**: By default, the thin pool writes zeros to each newly allocated block chunk before exposing it to a write operation. On single-tenant clusters, you can disable chunk zeroing to significantly reduce write amplification during initial data allocations.
 
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -221,9 +221,9 @@ When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass
   sudo lvextend --poolmetadatasize +1G <vgName>/<vgName>-thinpool
   ```
 
-### Choosing a VM Disk Bus
+### Choosing a Virtual Machine Disk Bus
 
-When attaching an LVM CSI PersistentVolumeClaim to a VirtualMachine, `virtio-scsi` (`bus: scsi`) generally performs better than the default `virtio-blk` (`bus: virtio`) for sustained-write workloads on thin-provisioned pools, particularly on RAID-backed storage. `virtio-scsi` supports multiple queues and has a more efficient DISCARD path.
+When attaching an LVM CSI PersistentVolumeClaim to a VirtualMachine, `virtio-scsi` (`bus: scsi`) generally outperforms the default `virtio-blk` (`bus: virtio`) for sustained-write workloads on thin-provisioned pools, particularly on RAID-backed storage. `virtio-scsi` performs better because it supports multiple queues and uses a more efficient DISCARD path.
 
 ### Coexistence with Longhorn v2 Block-Mode Disks
 
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -98,7 +98,19 @@ You can only use one type of local volume in each volume group. If necessary, cr
 
       ![](/img/v1.4/csi-driver-lvm/create-lvm-sc-04.png)
 
-1. Click **Save**.
+    - **Volume Group Type**: Select a type based on how the workload uses snapshots and how pool capacity is allocated. Harvester supports the following options:
+    
+        - **striped**: Best suited for workloads requiring direct, high-performance volume access distributed across the physical devices in the volume group. Each logical volume is fully allocated its requested capacity at provisioning time. Snapshots are created as independent logical volumes sized to match the source volume's maximum capacity. For example, a snapshot of a 100 GiB volume reserves an additional 100 GiB of volume group space upon creation, regardless of the actual quantity of data written to the source..
+
+        - **dm-thin**: Best suited for virtual machine workloads that frequently use snapshots or clones, and for environments that require over-provisioned pool capacity. Thin-provisioned volumes consume physical space only as blocks are written. Snapshots leverage true copy-on-write functionality at the thin-pool chunk level. For example, a fresh snapshot of a 100 GiB volume consumes practically zero pool capacity at creation, only growing as modified blocks accumulate over time.
+
+        :::tip
+
+        Select **dm-thin** if you expect to create regular snapshots or scheduled backups. This option is generally optimal for standard virtual machine workloads because of its efficient space utilization.
+
+        :::
+
+        ![](/img/v1.4/csi-driver-lvm/create-lvm-sc-04.png)
 
 1. On the **Storage** screen, verify that the StorageClass was created and the correct provisioner was set.
 
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -227,7 +227,7 @@ When attaching an LVM CSI PersistentVolumeClaim to a VirtualMachine, `virtio-scs
 
 ### Coexistence with Longhorn v2 Block-Mode Disks
 
-If the same node hosts a Longhorn v2 disk in block mode, the underlying device is held exclusively by the SPDK instance manager. Adding that device to the LVM `global_filter` prevents LVM's device scan from attempting to open it. Example, in `/etc/lvm/lvmlocal.conf`:
+If the same node hosts a Longhorn V2 disk in block mode, the underlying device is held exclusively by the SPDK Instance Manager. You can add this device to the LVM `global_filter` to exclude it from LVM device scans and prevent resource conflicts.
 
 ```
 devices {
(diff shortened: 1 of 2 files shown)


ibrokethecloud (dismissed):

abonillabeeche:
@jillian-maroket PING

jillian-maroket:
@abonillabeeche I'll work on this next week. Too many release deliverables.

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,93 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+- **Chunk size**: The chunk size determines the smallest unit of physical space that a thin pool allocates in response to a write. A write to a previously-unallocated region always allocates a full chunk, so small random writes to virgin space against a large chunk size cause severe write amplification (a 4 KiB write against a 16 MiB chunk allocates 16 MiB of pool space). When you install the harvester-csi-driver-lvm add-on version 0.4.0 or later, the default StorageClass sets `chunkSize: "1M"`. On the hardware-RAID-backed volume groups typical of Harvester nodes, `1M` matches the full-stripe width of common layouts (for example, four data disks at a 256 KiB strip yields a 1 MiB stripe), so each chunk allocation maps to whole stripes rather than partial ones. `1M` also keeps thin-pool metadata bounded and comfortably supports pools up to 256 TB. See the [Linux kernel dm-thin admin guide](https://docs.kernel.org/admin-guide/device-mapper/thin-provisioning.html) for background on chunk sizing.
```
jillian-maroket, line 200:
```suggestion
- **Chunk size**: The chunk size determines the smallest unit of physical space that a thin pool allocates in response to a write. Because a write to a previously unallocated region always provisions a full chunk, small random writes against a large chunk size cause severe write amplification (for example, a 4 KiB write against a 16 MiB chunk allocates 16 MiB of pool space).

  When you install the `harvester-csi-driver-lvm` add-on v0.4.0 or later, the default StorageClass uses `chunkSize: "1M"`. On the hardware RAID-backed volume groups typical of Harvester nodes, `1M` matches the full-stripe width of common layouts (for example, four data disks with a 256 KiB stripe size yield a 1 MiB stripe). This ensures each chunk allocation maps to whole stripes rather than partial ones. Additionally, a `1M` chunk size keeps thin-pool metadata bounded and comfortably supports pools up to 256 TB. For more information about chunk sizing, see the [Linux kernel dm-thin documentation](https://docs.kernel.org/admin-guide/device-mapper/thin-provisioning.html).
```

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,93 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+  :::warning
+
+  Chunk size cannot be changed after the thin pool is created. The value must be chosen before the first PersistentVolumeClaim is created against the StorageClass. Reducing chunk size on an existing pool requires evacuating all volumes, destroying the pool, recreating it with the new chunk size, and restoring the volumes.
+
+  :::
+
+  Align the chunk size to your RAID full-stripe width — that is, (number of data disks) × (RAID strip size). A chunk smaller than the full stripe (for example, a `512K` chunk on a 1 MiB stripe) forces every first-touch allocation into a partial-stripe **read-modify-write**. On parity RAID (5/6), partial-stripe writes also widen the write-hole window, so an unclean shutdown without a protected controller cache (BBU/FBWC) can leave a stripe with inconsistent parity. Choosing a chunk size equal to — or an exact multiple of — the full stripe avoids both problems; `1M` is the safe default because it aligns with the most common power-of-two RAID geometries.
```
jillian-maroket, line 208:
```suggestion
  Align the chunk size to your RAID full-stripe width: (number of data disks) × (RAID strip size). A chunk smaller than the full stripe (for example, a `512K` chunk on a 1 MiB stripe) forces every first-touch allocation into a partial-stripe **read-modify-write**. On parity RAID (5/6), partial-stripe writes also widen the write-hole window. This means that an unclean shutdown without a protected controller cache (BBU/FBWC) can leave a stripe with inconsistent parity. Choosing a chunk size equal to, or an exact multiple of, the full stripe avoids both issues. `1M` is the safe default because it aligns with the most common power-of-two RAID geometries.
```

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,93 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+  Override the default with the `chunkSize` parameter on the StorageClass. Common values:
+
+  | Value | When to use |
+  |---|---|
+  | `1M` | **Default** — matches the full-stripe width of common hardware-RAID layouts; general-purpose VM workloads |
+  | `512K` / `128K` | Only when the RAID full stripe is that size, or on non-RAID / single-disk volume groups where stripe alignment does not apply and minimizing snapshot copy-on-write cost is the priority |
+  | `2M` | Very large sequential-write pools whose RAID full stripe is 2 MiB |
```
jillian-maroket, line 216:
```suggestion
  Override the default with the `chunkSize` parameter on the StorageClass. Common values include the following:

  | Value | Target Environment |
  | :--- | :--- |
  | `1M` (Default) | Standard hardware RAID (1 MiB full stripe); general-purpose virtual machine workloads |
  | `512K` / `128K` | RAID arrays with matching full stripe size; non-RAID/single-disk volume groups where stripe alignment does not apply and minimizing snapshot copy-on-write overhead is primary |
  | `2M` | Large-stripe RAID arrays (2 MiB full stripe) with high-throughput sequential-write workloads |
```

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,93 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+  Never exceed `2M`. Larger chunks trigger disproportionate copy-on-write costs for snapshots, as documented in the [Red Hat Gluster admin guide](https://docs.redhat.com/en/documentation/red_hat_gluster_storage/3.5/html/administration_guide/chap-configuring_red_hat_storage_for_enhancing_performance). If you rely on LVM's built-in auto-selection instead of setting `chunkSize` explicitly, LVM chooses the chunk size based on the pool size to keep metadata bounded, which for multi-TB pools produces 8–16 MiB chunks — appropriate for metadata sizing but often not for random-write performance.
```
jillian-maroket, line 218:
```suggestion
  Do not exceed 2M. Larger chunk sizes trigger disproportionate copy-on-write overhead for snapshots, as documented in the [Red Hat Gluster admin guide](https://docs.redhat.com/en/documentation/red_hat_gluster_storage/3.5/html/administration_guide/chap-configuring_red_hat_storage_for_enhancing_performance).

  In addition, avoid relying on LVM's built-in auto-selection. To keep metadata bounded, LVM determines the chunk size based on the total pool size. This results in 8 to 16 MiB chunks for multi-terabyte pools, which is appropriate for metadata sizing but can severely degrade random-write performance. Always set `chunkSize` explicitly in the StorageClass.
```

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,93 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+  Set `zeroBlocks: "false"` on the StorageClass at pool creation time, or apply it to an existing pool with:
+
+  ```
+  sudo lvchange --zero n <vgName>/<vgName>-thinpool
+  ```
+
+  You can fully reverse the change by running the command with `--zero y`.
+
+- **Pool metadata size**: When the thin pool is created, the driver sizes its metadata logical volume based on the `poolMetadataSize` StorageClass parameter (default `16G`, sufficient for pools up to 256 TB at the default 1 MiB chunk size). A larger chunk size needs less metadata to address the same capacity, and a smaller chunk size needs more: metadata bytes ≈ 64 × (pool_size ÷ chunk_size). If you did not set the parameter, or if the pool grew significantly after creation, extend the metadata volume proactively to prevent the pool from running out of space and becoming unresponsive later.
```
jillian-maroket, line 237:
```suggestion
- **Pool metadata size**: When the thin pool is created, the driver sizes its metadata logical volume based on the `poolMetadataSize` StorageClass parameter. The default size is `16G`, which supports pools up to 256 TB at the default 1 MiB chunk size. Larger chunk sizes require less metadata to address the same capacity, while smaller chunk sizes require more (`metadata_bytes ≈ (pool_size ÷ chunk_size) × 64`). If you do not set this parameter, or if the pool grows significantly after creation, proactively extend the metadata volume to prevent the pool from exhausting metadata space and becoming unresponsive later.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -104,6 +104,16 @@ You can only use one type of local volume in each volume group. If necessary, cr
 
     ![](/img/v1.4/csi-driver-lvm/create-lvm-sc-05.png)
 
+### Considerations: `striped` vs `dm-thin`
+
+Both volume group types are fully supported. The choice depends on how the workload will use snapshots and how the pool capacity will be shared.
```
jillian-maroket, line 109:
```suggestion
Harvester supports both volume group types. The choice depends on how the workload will use snapshots and how the pool capacity will be shared.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -104,6 +104,16 @@ You can only use one type of local volume in each volume group. If necessary, cr
...
+
+- **`striped`** is a good fit for workloads that mostly need direct volume performance across the physical devices in the volume group. Each logical volume gets its full requested capacity at provision time. Snapshots are provisioned as separate LVs sized to the origin's full capacity — a snapshot of a 100 GiB volume reserves an additional 100 GiB of volume group space at creation, regardless of how much data has actually been written to the origin.
```
jillian-maroket, line 111:
```suggestion
- **striped**: Best suited for workloads requiring direct, high-performance volume access distributed across the physical devices in the volume group. Each logical volume is fully allocated its requested capacity at provisioning time. Snapshots are created as independent logical volumes sized to match the source volume's maximum capacity. For example, a snapshot of a 100 GiB volume reserves an additional 100 GiB of volume group space upon creation, regardless of the actual quantity of data written to the source.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -104,6 +104,16 @@ You can only use one type of local volume in each volume group. If necessary, cr
...
+
+- **`dm-thin`** is a good fit for virtual machine workloads that take snapshots or clones, and for environments that want to over-provision pool capacity. Thin volumes consume physical space only as blocks are written to them, and snapshots are true copy-on-write at the thin-pool chunk level — a fresh snapshot of a 100 GiB volume adds effectively zero pool capacity at creation and only grows as changed blocks accumulate.
```
jillian-maroket, line 113:
```suggestion
- **dm-thin**: Best suited for virtual machine workloads that frequently use snapshots or clones, and for environments that require over-provisioned pool capacity. Thin-provisioned volumes consume physical space only as blocks are written. Snapshots leverage true copy-on-write functionality at the thin-pool chunk level. For example, a fresh snapshot of a 100 GiB volume consumes practically zero pool capacity at creation, only growing as modified blocks accumulate over time.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -104,6 +104,16 @@ You can only use one type of local volume in each volume group. If necessary, cr
...
+
+If regular snapshots are expected, `dm-thin` is generally the right choice for VM workloads.
```
jillian-maroket, line 115:
```suggestion
:::tip

Select **dm-thin** if you expect to create regular snapshots or scheduled backups. This option is generally optimal for standard virtual machine workloads because of its efficient space utilization.

:::

```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +187,45 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
 
-:::
\ No newline at end of file
+:::
+
+## Additional Notes
+
+### Tuning the `dm-thin` Pool
+
+When the first PersistentVolumeClaim is created against a `dm-thin` StorageClass, the driver creates an LVM thin-pool named `<vgName>-thinpool` at `-l 90%FREE` of the volume group. Two settings are worth knowing about:
```
jillian-maroket, line 196:
```suggestion
When the first PersistentVolumeClaim (PVC) is created against a `dm-thin` StorageClass, the driver creates an LVM thin pool named `<vgName>-thinpool` using `-l 90%FREE` (allocating 90% of the volume group's remaining free space). Consider tuning the following settings based on your workload demands:
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +187,45 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+- **Chunk zeroing.** By default, the pool writes zeros to each newly-allocated chunk before handing it to the writer. On single-tenant clusters this can be disabled to reduce write amplification on first-touch allocations:
```
jillian-maroket, line 198:
```suggestion
- **Chunk zeroing**: By default, the thin pool writes zeros to each newly allocated chunk before exposing it to a write operation. On single-tenant clusters, you can disable chunk zeroing to significantly reduce write amplification during initial data allocations.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +187,45 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+  ```
+  sudo lvchange --zero n <vgName>/<vgName>-thinpool
+  ```
+
+  The change is fully reversible with `--zero y`.
+
+- **Pool metadata size.** LVM auto-sizes the thin-pool metadata LV at pool creation. For pools that will hold many snapshots or many thin volumes over time, extending the metadata proactively avoids running short later:
```
jillian-maroket, line 206:
```suggestion
- **Pool metadata size**: When the thin pool is created, LVM automatically sizes its metadata logical volume. Consider extending this volume proactively if you expect the pool to store a large number of snapshots or thin volumes over time. Doing so prevents the pool from running out of space and becoming unresponsive later.
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +187,45 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+  sudo lvextend --poolmetadatasize +1G <vgName>/<vgName>-thinpool
+  ```
+
+### Choosing a VM Disk Bus
+
+When attaching an LVM CSI PersistentVolumeClaim to a VirtualMachine, `virtio-scsi` (`bus: scsi`) generally performs better than the default `virtio-blk` (`bus: virtio`) for sustained-write workloads on thin-provisioned pools, particularly on RAID-backed storage. `virtio-scsi` supports multiple queues and has a more efficient DISCARD path.
+
+### Coexistence with Longhorn v2 Block-Mode Disks
+
+If the same node hosts a Longhorn v2 disk in block mode, the underlying device is held exclusively by the SPDK instance manager. Adding that device to the LVM `global_filter` prevents LVM's device scan from attempting to open it. Example, in `/etc/lvm/lvmlocal.conf`:
```
jillian-maroket, line 218:
```suggestion
### Coexistence with Longhorn V2 Block Mode Disks

If the same node hosts a Longhorn V2 disk in block mode, the underlying device is held exclusively by the SPDK Instance Manager. You can add this device to the LVM `global_filter` to exclude it from LVM device scans and prevent resource conflicts.

Example (`/etc/lvm/lvmlocal.conf`):
```

versioned_docs/version-v1.8/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +187,45 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+    global_filter = [
+      "r|/dev/loop.*|",
+      "r|/dev/disk/by-path/.*longhorn.*|",
+      "r|/dev/mapper/pvc-.*|",
+      "r|/dev/disk/by-id/wwn-<lhv2-disk-wwn>|",
+    ]
+}
+```
+
+On Harvester's immutable OS, persist the change through an `/oem/*.yaml` cloud-config file so it survives reboots. For background, see [harvester/harvester#11098](https://github.com/harvester/harvester/issues/11098).
```
jillian-maroket, line 231:
```suggestion
Because Harvester's operating system is immutable, you must persist this change through an `/oem/*.yaml` cloud-config file to ensure it survives system reboots and upgrades.For more information, see issue [#11098](https://github.com/harvester/harvester/issues/11098).
```

jillian-maroket (commented):
Review done

abonillabeeche:
All 13 pending suggestions from the 2026-08-11 review are applied in 839f918 — accepted as written except where a suggestion's context had already shifted slightly from later commits, in which case I matched the surrounding text to the intent of the suggestion. Covers both the main doc's chunk-size/pool-metadata rewording and the v1.8 versioned copy's striped-vs-dm-thin section and Longhorn coexistence heading.

@jillian-maroket please take another look when you have a chance.

jillian-maroket (dismissed):

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -98,7 +98,19 @@ You can only use one type of local volume in each volume group. If necessary, cr
 
       ![](/img/v1.4/csi-driver-lvm/create-lvm-sc-04.png)
 
-1. Click **Save**.
+    - **Volume Group Type**: Select a type based on how the workload uses snapshots and how pool capacity is allocated. Harvester supports the following options:
+    
+        - **striped**: Best suited for workloads requiring direct, high-performance volume access distributed across the physical devices in the volume group. Each logical volume is fully allocated its requested capacity at provisioning time. Snapshots are created as independent logical volumes sized to match the source volume's maximum capacity. For example, a snapshot of a 100 GiB volume reserves an additional 100 GiB of volume group space upon creation, regardless of the actual quantity of data written to the source..
```
WebberHuang1118, line 103: Could we avoid recommending `striped` for new workloads here? [harvester/harvester#11334](https://github.com/harvester/harvester/issues/11334) tracks its deprecation before GA and identifies `dm-thin` as the supported type. Please label `striped` as deprecated and outside GA scope, recommend `dm-thin` for new configurations, and add compatibility or migration guidance for existing pre-GA striped volumes.

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,97 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+  Verify the effective chunk size of a live pool:
+
+  ```
+  sudo lvs -o vg_name,lv_name,chunk_size,zero <vgName>/<vgName>-thinpool
+  sudo dmsetup table <vgName>-<vgName>--thinpool-tpool
+  ```
+
+- **Chunk zeroing**: By default, the thin pool writes zeros to each newly allocated block chunk before exposing it to a write operation. On single-tenant clusters, you can disable chunk zeroing to significantly reduce write amplification during initial data allocations.
+
+  Set `zeroBlocks: "false"` on the StorageClass at pool creation time, or apply it to an existing pool with:
```
WebberHuang1118, line 233: Please remove the instructions to run `lvchange` and `lvextend` on a live CSI-managed thin pool. Direct pool changes can conflict with attached volumes and driver operations. Any manual offline procedure should require that all volumes are detached and no corresponding dm devices exist before changing the pool.

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,97 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+  ```yaml
+  apiVersion: storage.k8s.io/v1
+  kind: StorageClass
+  metadata:
+    name: harvester-lvm-thin
+  provisioner: lvm.driver.harvesterhci.io
+  parameters:
+    type: dm-thin
+    vgName: vmvg
+    chunkSize: "1M"            # Applied on first-time pool creation only.
```
WebberHuang1118, line 258: These settings belong to the thin pool shared by all LVs in the volume group, not to an individual StorageClass. The first pool creation determines them, so presenting them as per-StorageClass tuning knobs is misleading and can lead to conflicting configurations. Please remove this example and recommendation; [#11335](https://github.com/harvester/harvester/issues/11335) tracks a dedicated pool-level configuration and status resource.

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,97 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+    poolMetadataSize: "16G"    # Applied on first-time pool creation only.
+    zeroBlocks: "false"        # Applied on first-time pool creation only.
+  reclaimPolicy: Delete
+  volumeBindingMode: WaitForFirstConsumer
+  allowVolumeExpansion: true
+  ```
+
+### Choosing a Virtual Machine Disk Bus
+
+When attaching an LVM CSI PersistentVolumeClaim to a VirtualMachine, `virtio-scsi` (`bus: scsi`) generally outperforms the default `virtio-blk` (`bus: virtio`) for sustained-write workloads on thin-provisioned pools, particularly on RAID-backed storage. `virtio-scsi` performs better because it supports multiple queues and uses a more efficient DISCARD path.
```
WebberHuang1118, line 268: Could we remove this general performance recommendation or support it with a reproducible Harvester LVM CSI benchmark? [QEMU recommends `virtio-blk` for performance-critical use cases](https://www.qemu.org/2021/01/19/virtio-blk-scsi-configuration/), and [KubeVirt supports multiqueue on `virtio-blk` and enables discard passthrough by default](https://kubevirt.io/user-guide/storage/disks_and_volumes/). Our Windows VM guide also reports better `virtio-blk` results on LVM CSI. We can describe SCSI-specific feature needs without claiming it generally writes faster.

docs/advanced/addons/lvm-local-storage.md
```diff
@@ -177,4 +189,97 @@ You can also create a new virtual machine with the volume of the LVM StorageClas
...
+
+### Coexistence with Longhorn v2 Block-Mode Disks
+
+If the same node hosts a Longhorn V2 disk in block mode, the underlying device is held exclusively by the SPDK Instance Manager. You can add this device to the LVM `global_filter` to exclude it from LVM device scans and prevent resource conflicts.
```
WebberHuang1118, line 272: Could we leave this coexistence guidance with @tserong as part of [#11098](https://github.com/harvester/harvester/issues/11098)? The issue is still under investigation and discusses a possible devices-file fix, while this section presents the `global_filter` workaround as a general user procedure. Please have the issue owner validate the supported guidance before publishing it.

WebberHuang1118 (commented):
Please address the inline documentation concerns before publishing this guidance, thanks.

docs/advanced/addons/lvm-local-storage.md
abonillabeeche: Done in 7e71305 — this also collapses the duplicated **Volume Group Type** bullet that this PR had accidentally introduced alongside the existing one.

`striped` is now explicitly labeled deprecated and outside the GA support scope, `dm-thin` is called out as the supported type and leads the list, and the tip that recommended `dm-thin` only for snapshot-heavy workloads is gone. I added the user-facing rationale you asked for in #11334 (no storage-efficient snapshot/clone/restore; restore copies the full source capacity, so capacity/IO/timeout cost scales with volume size) plus compatibility and migration guidance: existing pre-GA striped volumes keep working, there is no in-place conversion, and the migration path is a new dm-thin volume group → StorageClass → volume, attach alongside, copy at the guest/application level, decommission after verification.

I kept the migration steps deliberately generic rather than naming a specific Harvester mechanism, since backup is still unsupported for LVM CSI. If you would rather point at a concrete procedure, tell me which one is considered supported and I will write it up.

docs/advanced/addons/lvm-local-storage.md
abonillabeeche: Removed in 7e71305. Both the `lvchange --zero n` and the `lvextend --poolmetadatasize` commands are gone.

`zeroBlocks` is now documented as something you set on the StorageClass before the pool is created, and the metadata bullet now says to size the volume up front for the capacity the pool is expected to reach, noting that a pool which exhausts metadata space becomes unresponsive and can only be corrected offline.

I also added a warning at the top of the tuning section stating the general rule: do not run `lvchange` / `lvextend` / `lvconvert` against a CSI-managed thin pool while it is in use, and any manual procedure must be performed offline with all volumes in the pool detached and no corresponding device-mapper devices present on the node.

The only LVM commands left in the section are the read-only `lvs` / `dmsetup table` used to verify the effective chunk size. Happy to drop those too if you would prefer.

docs/advanced/addons/lvm-local-storage.md
abonillabeeche: Agreed — the example is removed in 7e71305.

In its place the tuning section now opens with an info admonition stating that these settings belong to the thin pool shared by every LV in the volume group and are not properties of an individual StorageClass; that the StorageClass which first triggers pool creation determines chunk size, metadata size and zeroing; that the same parameters on StorageClasses created later against the same volume group have no effect; and that defining several StorageClasses with different values therefore yields a configuration that does not match what is applied. It links #11335 for the dedicated pool-level configuration and status resource.

I also reworded the two places in the chunk-size bullet that read as per-StorageClass tuning, so they now refer to the StorageClass that first provisions into the volume group.

docs/advanced/addons/lvm-local-storage.md
abonillabeeche: You are right, and our own data agrees with you — the Windows VM benchmark on LVM CSI measured `virtio-blk` roughly 2x faster than `virtio-scsi` on random-write and mixed workloads, so the claim as written was backwards. Rewritten in 7e71305.

The section now advises retaining the default `virtio-blk` unless a SCSI-specific capability is required, citing the QEMU and KubeVirt references you linked, and describes when `bus: scsi` is actually needed (SCSI persistent reservations for clustered guests such as WSFC, SCSI passthrough, and attaching more disks than the guest virtio-blk device budget allows) without any general performance claim.

docs/advanced/addons/lvm-local-storage.md
abonillabeeche: Makes sense — the `global_filter` block is removed in 7e71305, so nothing prescriptive is published ahead of the investigation.

I kept a short symptom-level note so that a user who hits this has a breadcrumb: Longhorn V2 block-mode holds the device exclusively via the SPDK Instance Manager, LVM commands against volume groups on that node can block indefinitely, and LVM CSI provisioning and detach operations stall as a result. It then states that no supported workaround is currently available and links #11098.

@tserong, over to you — if the devices-file approach lands, or if there is interim guidance you consider supported, let me know and I will document it here.

PR #1098, opened 2026-08-05
Add hint for /oem/harvester.config

votdev:
I think it makes sense and would be helpful to include this information in the documentation.

grafik

### DIFF
--- a/docs/install/update-harvester-configuration.md
+++ b/docs/install/update-harvester-configuration.md
@@ -24,6 +24,12 @@ If you upgrade from a version before `v1.1.2`, the `cloud-init` file in examples
 
 :::
 
+:::info important
+
+After installation, Harvester does not re-read or reconcile changes made to `/oem/harvester.config`.
+
+:::
+
 ## Password of user `rancher`
 
 ### Runtime change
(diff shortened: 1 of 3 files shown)


docs/install/update-harvester-configuration.md
```diff
@@ -24,6 +24,12 @@ If you upgrade from a version before `v1.1.2`, the `cloud-init` file in examples
 
 :::
 
+:::note
+
+Once Harvester is installed, changes to `/oem/harvester.config` are **never** re-read, parsed, or reconciled by Harvester, even if updated or customized.
+
+:::
```
jillian-maroket:
```suggestion
:::info important

After installation, Harvester does not re-read or reconcile changes made to `/oem/harvester.config`.

:::
```

jillian-maroket (commented):
Changed the admonition type and tightened the wording

tserong:
Good call out.

I should mention though that there's one exception to this. If the user runs `harvester-installer generate-network-config` it will re-create the NM connection profiles based on the contents of `/oem/harvester.config` - see https://docs.harvesterhci.io/v1.8/upgrade/v1-6-x-to-v1-7-x#migration-from-wicked-to-networkmanager. That's a way of re-setting your NM connection profiles back to what they were originally if you happen to screw them up somehow. But, this is not something that Harvester will ever automatically do.

jillian-maroket (approved):

PR #1104, opened 2026-08-12
docs: correct the LVM yaml file path

Vicente-Cheng:
- we should use the corresponding versioned path

#### Problem:
Update the LVM yaml file path

#### Solution:
Update the LVM yaml file path

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10754


jillian-maroket (approved):

WebberHuang1118 (approved):
LGTM, thanks.

PR #1107, opened 2026-08-19
docs(lvm): document encryption at rest for LVM volumes

abonillabeeche:
## What

Documents **encryption at rest for LVM volumes** on the LVM local storage add-on page
(`docs/advanced/addons/lvm-local-storage.md`), covering:

- Creating the `CRYPTO_KEY_*` encryption secret (same convention as Longhorn).
- Creating an encrypted `dm-thin` StorageClass from the UI (**Volume Encryption** toggle
  + secret) and the equivalent YAML.
- The `cloneStrategy: copy` requirement so VM-image clones are written through the
  encryption layer.
- That VM images, volumes, VMs, snapshots, and backups/restores inherit encryption by
  using the encrypted StorageClass.

## Related

- Tracking issue: harvester/harvester#11442
- Driver support: harvester/csi-driver-lvm#67
- Dashboard UI (Volume Encryption toggle for the LVM provisioner): harvester/harvester-ui-extension#1121

## Notes

- Targets the next (main) docs version. Happy to backport to `versioned_docs/version-v1.x`
  once the driver + UI land in a release.
- Draft: screenshots for the new UI steps can be added after the dashboard PR merges.

### DIFF
--- a/docs/advanced/addons/lvm-local-storage.md
+++ b/docs/advanced/addons/lvm-local-storage.md
@@ -118,7 +118,9 @@ Encryption is an opt-in property of the StorageClass. Existing (unencrypted) Sto
 
 ### Creating an Encryption Secret
 
-Create a `Secret` that holds the encryption passphrase in the `CRYPTO_KEY_VALUE` field. The remaining `CRYPTO_KEY_*` fields are optional and default to `aes-xts-plain64` / `sha256` / `256` / `argon2i` when omitted.
+Create a `Secret` that holds the encryption passphrase in the `CRYPTO_KEY_VALUE` field. Include all of the fields shown below: the CSI driver applies the defaults `aes-xts-plain64` / `sha256` / `256` / `argon2i` when the tuning fields are omitted, but Harvester rejects a StorageClass whose encryption secret has a missing or empty field.
+
+The secret must exist before you create the StorageClass that references it.
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


abonillabeeche:
Updated this PR (`aab00c2`) after validating the feature end to end on a Harvester cluster (RKE2 v1.35.6, dm-thin) with harvester/csi-driver-lvm#67. Four corrections, all of them things a user would have hit by following the previous text:

1. **Removed the `${pvc.name}` / `${pvc.namespace}` guidance.** Templating works on upstream Kubernetes, but Harvester resolves the secret reference literally when the StorageClass is admitted, so a templated class is rejected (`secret ${pvc.namespace}/${pvc.name}-luks not found`). The text now says to create one encrypted StorageClass per encryption secret, and notes that the secret must exist beforehand.
2. **Added the `node-expand-secret-*` references to the example.** Expanding an encrypted volume resizes the dm-crypt mapper first, which needs the passphrase; without these two parameters expansion fails with `InvalidArgument`.
3. **The `CRYPTO_KEY_*` fields are not optional on Harvester.** The driver applies defaults when they are omitted, but the StorageClass webhook rejects a secret with a missing or empty field (`… is not a valid encryption secret, missing field: CRYPTO_KEY_SIZE`).
4. **Added a short "Snapshots, Clones, and Restores of Encrypted Volumes" section.** A snapshot or clone is a block-level copy, so encryption state cannot be changed by a restore (both mismatch directions are now rejected by the driver), and a volume restored from an encrypted snapshot needs its *source's* passphrase — restoring through a class that references a different secret provisions but then fails to attach. This matches the behaviour validated in harvester/csi-driver-lvm#67 and tracked in harvester/harvester#11557.

DCO is green on the new commit.

PR #1122, opened 2026-09-02
Mention `Image Volume Size Check` in 1.9.x to 1.10.x upgrade

votdev:
The following changes are applied by the PR:

- add information about the `Image Volume Size Check` that is performed during the upgrade from v1.9.x to v1.10.0.
- Modify the `sidebar_position` in the front matter of all MD files below `docs/upgrade`.
- The MD file for the upgrade `v1.8.x to v1.9.0` is added as an empty template. Should be added to `main` branch soonish when v1.9.0 is released.

Related to: https://github.com/harvester/harvester/issues/10351


docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,42 @@
...
+  <link rel="canonical" href="https://docs.harvesterhci.io/v1.10/upgrade/v1-9-x-to-v1-10-x"/>
+</head>
+
+## General Information
+
+The **Upgrade** button on the **Dashboard** screen becomes selectable whenever a new Harvester version that you can upgrade to becomes available. For more information, see [Start an upgrade](./automatic.md#start-an-upgrade).
+
+### Image Volume Size Check
+
+Prior to v1.10.0, Harvester did not validate that a volume created from a VM image was at least as large as the image's virtual size. A volume smaller than its source image can lead to data integrity issues once the guest OS writes past the shrunk backing store. Starting with v1.10.0, Harvester rejects any new volume request that violates this constraint, but volumes created before the upgrade may still be affected.
```
WebberHuang1118: The image volume check should already be merged and included starting from v1.9.0, right?

WebberHuang1118 (commented):
Should we provide a way to check the image volume size before starting the upgrade, similar to the pre-check script described [here](https://docs.harvesterhci.io/v1.9/upgrade/index#start-an-upgrade)?

votdev:
> Should we provide a way to check the image volume size before starting the upgrade, similar to the pre-check script described [here](https://docs.harvesterhci.io/v1.9/upgrade/index#start-an-upgrade)?

Yes, we came to the conclusion [here](https://github.com/harvester/harvester/pull/11315#issuecomment-5537630553) to add a separate pre-flight check script.

docs/upgrade/v1-9-x-to-v1-10-x.md
votdev: Since harvester/upgrade-helpers isn't tied to any specific releases, we can include the check in every version of Harvester that's still supported after https://github.com/harvester/upgrade-helpers/pull/53 has been merged.

However, I would explicitly address this on the `Upgrade from v1.9.x to v1.10.x` page, since the webhook validator is introduced in v1.10.0. From that point on, the check makes perfect sense and should therefore be explicitly mentioned. For previous versions, the check will be available once the [PR](https://github.com/harvester/upgrade-helpers/pull/53) is merged into `harvester/upgrade-helpers`. However, since we rarely mention checks explicitly in the documentation, we should consider updating the [README](https://github.com/harvester/upgrade-helpers/blob/main/pre-check/v1.x/README.md) for `harvester/upgrade-helpers`. The file hasn’t been updated in a long time and no longer mentions the checks that have been added recently.

WebberHuang1118 (dismissed):
LGTM, thanks.

docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,48 @@
...
+
+## General Information
+
+The **Upgrade** button on the **Dashboard** screen becomes selectable whenever a new Harvester version that you can upgrade to becomes available. For more information, see [Start an upgrade](./automatic.md#start-an-upgrade).
+
+### Volume Size Validation for VM Images
+
+Starting with v1.10.0, Harvester validates that a volume created from a VM image is at least as large as the virtual size of that image, rounded up to whole GiB, and rejects any volume request that violates this constraint. Earlier versions did not perform this validation, so a volume could end up smaller than its source image. This most commonly happened when a golden image was replaced by a larger one after volumes had already been created from it. Because the guest operating system sizes its filesystem according to the image, writing past the end of an undersized volume can corrupt the guest filesystem.
+
+The validation applies to volume requests only. Volumes that already exist when you upgrade are not modified, so any volume that is currently undersized remains undersized until you expand it.
```
jillian-maroket, line 19:
```suggestion
Starting with v1.10.0, Harvester validates that any new volume created from a virtual machine image is at least as large as the virtual size of that image (rounded up to the nearest whole GiB). Harvester rejects volume requests that violate this constraint.

This validation applies exclusively to new volume requests. Existing volumes are not modified during an upgrade, which means that any previously undersized volume remains undersized until expanded manually.

#### Root Cause and Risk

Earlier Harvester versions did not enforce volume size validation, allowing volumes to be created smaller than their source images. This issue typically occurred when a golden image was replaced by a larger version after volumes had already been created from it.

Because the guest operating system provisions its filesystem based on the virtual image size, writing data past the physical end of an undersized volume can cause guest filesystem corruption.
```

docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,48 @@
...
+
+To identify such volumes before you start the upgrade, run the [pre-check script](https://github.com/harvester/upgrade-helpers/tree/main/pre-check) on a Harvester control-plane node. The script is maintained independently of Harvester releases, so always download the current version:
```
jillian-maroket, line 21:
```suggestion
#### Identifying Undersized Volumes

Before starting the upgrade, run the [pre-check script](https://github.com/harvester/upgrade-helpers/tree/main/pre-check) on a Harvester management node to identify affected volumes.

Because the script is updated independently of Harvester releases, ensure that you download the latest version:
```

docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+# Download the script
+$ curl -o /tmp/check.sh https://raw.githubusercontent.com/harvester/upgrade-helpers/refs/heads/main/pre-check/v1.x/check.sh && chmod +x /tmp/check.sh
+
+# Run the checks
+$ /tmp/check.sh
+```
+
+Its **Image Volume Size** check compares the requested size of every PersistentVolumeClaim (PVC) that was created from a VM image against the virtual size of that image and reports each affected volume:
```
jillian-maroket, line 31:
```suggestion
The **Image Volume Size** check compares the requested size of every PersistentVolumeClaim (PVC) created from an image against that image's virtual size. If undersized volumes are identified, the script displays output similar to the following:
```

docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,48 @@
...
+```
+Starting Image Volume Size check...
+Found volumes that are smaller than the virtual size of the VM image they were created from:
+  default/test-pvc: 5Gi is smaller than the required minimum of 10Gi (source image default/test-image)
+Once the guest writes past the end of such a volume, the guest filesystem can be corrupted.
+Expand each of the volumes listed above to at least the required minimum size. If the storage class does not support online expansion, shut down the virtual machine using the volume first.
+Image-Volume-Size Test: Failed
+```
+
+The check is read-only. It only reports the affected volumes and never resizes anything. A finding is counted as a failed check, so the script also lists it in the summary at the end of the run and exits with a non-zero status:
```
jillian-maroket, line 42:
```suggestion
This check is strictly read-only and does not modify any resources. Any identified undersized volume causes the check to fail, resulting in a non-zero exit code and a summary alert:
```

docs/upgrade/v1-9-x-to-v1-10-x.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+WARN: There are 1 failing checks: Image-Volume-Size
+```
+
+For each volume reported as undersized, increase `spec.resources.requests.storage` to at least the required minimum size before you start the upgrade. If the storage class does not support online expansion, shut down the virtual machine using the volume first.
```
jillian-maroket, line 48:
```suggestion
#### Remediation

To resolve this issue before upgrading, perform the following steps for each identified undersized volume:

1. Check if the underlying StorageClass supports online volume expansion. If online expansion is not supported, shut down the virtual machine that is using the volume.

1. Edit the PVC manifest and increase the value of `spec.resources.requests.storage` to at least the required minimum size reported by the script.

1. Re-run `/tmp/check.sh` to confirm that all volume size checks pass.
```

docs/vm/create-vm.md
```diff
@@ -164,7 +164,7 @@ If you are using external storage, ensure that the correct **StorageClass** and
 
 :::info important
 
-When creating volumes from a VM image, ensure that the volume size is greater than or equal to the image size. The volume may become corrupted if the configured volume size is less than the size of the underlying image. This is particularly important for qcow2 images because the virtual size is typically greater than the physical size.
+Starting with Harvester v1.10.0, Harvester strictly **validates and enforces** that any volume created or updated from a VM image must have a size greater than or equal to the virtual size of the source image. The validating webhook will reject any VM disk/volume requests that violate this constraint to prevent potential guest data corruption.
```
jillian-maroket, line 167:
```suggestion
Starting with Harvester v1.10.0, Harvester validates that any volume created or updated from a virtual machine image is at least as large as the virtual size of the source image. The validating webhook rejects virtual machine disk or volume requests that violate this constraint to prevent guest filesystem corruption.
```

docs/volume/create-volume.md
```diff
@@ -135,7 +135,7 @@ resource "harvester_volume" "empty-volume" {
 
 :::info important
 
-When creating volumes from a VM image, ensure that the volume size is greater than or equal to the image size. The volume may become corrupted if the configured volume size is less than the size of the underlying image. This is particularly important for qcow2 images because the virtual size is typically greater than the physical size.
+Starting with Harvester v1.10.0, Harvester strictly **validates and enforces** that any volume created or updated from a VM image must have a size greater than or equal to the virtual size of the source image. The validating webhook will reject any PVC creation or update requests that violate this constraint to prevent potential guest data corruption.
```
jillian-maroket, line 138:
```suggestion
Starting with Harvester v1.10.0, Harvester validates that any volume created or updated from a virtual machine image is at least as large as the virtual size of the source image. The validating webhook rejects PVC creation or update requests that violate this constraint to prevent guest filesystem corruption.
```

jillian-maroket (commented):
Review done

jillian-maroket (approved):

WebberHuang1118 (approved):

PR #1124, opened 2026-09-08
Fix link to pre-check script in upgrade documentation

saurabhsadhalesuse:
Updated the link to the pre-check script for clarity.

#### Problem:
The link always requires you to click on v1.x and then check.sh
#### Solution:
I have updated the link to the script directly.


starbops (approved):
LGTM, thank you for the fix.

jillian-maroket (approved):

PR #1125, opened 2026-09-09
docs: clarify StorageClass sync note across all versioned docs

rajeshkio:
#### Problem:
The documentation for VM image sync during cluster restore states that the controller automatically syncs virtual machine images to the new cluster, but doesn't mention that each image's associated StorageClass is also synced as part of this process. This omission could lead users to assume StorageClasses need to be manually recreated after a cluster restore.

#### Solution:
Updated the description to clarify that StorageClasses associated with each virtual machine image are also synced automatically to the new cluster, alongside the image itself. Applied consistently across the current docs and all affected versioned doc snapshots (v1.4–v1.8).

#### Related Issue(s):
https://github.com/harvester/harvester/issues/11613

### DIFF
--- a/docs/vm/backup-restore.md
+++ b/docs/vm/backup-restore.md
@@ -143,7 +143,7 @@ Users can now restore a new VM on another cluster by leveraging the VM metadata 
 
 #### Prerequisites
 
-- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, except when a virtual machine image with the same name or display name already exists on the new cluster.
+- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, along with their associated StorageClasses, except when a virtual machine image with the same name or display name already exists on the new cluster.
 
 - Earlier than v1.4.0: You must upload and configure the virtual machine images on the new cluster. Ensure that the image names and configuration are identical so that the virtual machines can be restored. For more information, see [Upload the same VM images to a new cluster](#upload-the-same-vm-images-to-a-new-cluster).
 
--- a/versioned_docs/version-v1.4/vm/backup-restore.md
+++ b/versioned_docs/version-v1.4/vm/backup-restore.md
@@ -157,7 +157,7 @@ Users can now restore a new VM on another cluster by leveraging the VM metadata 
 
 #### Prerequisites
 
-- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, except when a virtual machine image with the same name or display name already exists on the new cluster.
+- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, along with their associated StorageClasses, except when a virtual machine image with the same name or display name already exists on the new cluster.
 
 - Earlier than v1.4.0: You must upload and configure the virtual machine images on the new cluster. Ensure that the image names and configuration are identical so that the virtual machines can be restored. For more information, see [Upload the same VM images to a new cluster](#upload-the-same-vm-images-to-a-new-cluster).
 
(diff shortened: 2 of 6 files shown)


jillian-maroket (approved):

docs/vm/backup-restore.md
```diff
@@ -143,7 +143,7 @@ Users can now restore a new VM on another cluster by leveraging the VM metadata
 
 #### Prerequisites
 
-- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, except when a virtual machine image with the same name or display name already exists on the new cluster.
+- v1.4.0 and later: The controller automatically syncs the virtual machine images with the new cluster, along with their associated StorageClasses, except when a virtual machine image with the same name or display name already exists on the new cluster.
```
Vicente-Cheng, line 146: Not all StorageClass would be synced.
IIUC, only the backing image (vmimage) would be synced.

cc @WebberHuang1118, please correct me if I am wrong.

rajeshkio:
VMI associated storageclass are the one auto-created when we import an image right? The document update is meant to rely that information only.

docs/vm/backup-restore.md
WebberHuang1118: Yes, you are right, only the sc related to V1 VMImage. We should provide more detailed context instead of saying that the associated StorageClasses are synced:

- Currently, VM Backup and Restore supports only Longhorn v1 volumes.
- Harvester can automatically back up VMImages to the remote backup target. However, both the backup and restore clusters must be running Harvester v1.4.0 or later. See the [prerequisites](https://docs.harvesterhci.io/v1.9/vm/backup-restore/#prerequisites). In addition, the two clusters should not have VMImages with duplicate names or display names.
- The backup target **Refresh Interval** setting affects how frequently Harvester discovers existing VMBackups and VMImages from the remote backup target. See [Configure Backup Target](https://docs.harvesterhci.io/v1.9/vm/backup-restore/#configure-backup-target).

WebberHuang1118 (commented):
Thanks  for the improvement, please check https://github.com/harvester/docs/pull/1125/changes#r4005984249, thanks.

docs/vm/backup-restore.md
rajeshkio: I am thinking of this rewrite, does it looks okay @WebberHuang1118 

v1.4.0 and later: Harvester can automatically back up VMImages to the remote backup target. Both the backup and restore clusters must be running Harvester v1.4.0 or later, and should not have VMImages with duplicate names or display names; sync is skipped for a VMImage if one already exists with the same name or display name. The controller also syncs each VMImage's associated StorageClass. The backup target Refresh Interval setting affects how frequently Harvester discovers existing VMBackups and VMImages from the remote backup target. See Configure Backup Target.

Your first point is already present in the doc

"Backup support is currently limited to Longhorn V1 Data Engine volumes. Harvester is unable to create backups of volumes in external storage."

WebberHuang1118: LGTM, thanks

jillian-maroket:
```suggestion
- v1.4.0 and later: Harvester can automatically sync and back up virtual machine images to a remote backup target. Both source and target clusters must run v1.4.0 or later and must not contain images with identical names or display names. Harvester skips syncing an image if one with the same name or display name already exists on the target cluster. When syncing an image, the controller also automatically syncs its associated StorageClass.

  The Refresh Interval setting on the backup target determines how frequently Harvester scans the remote target for new or updated backups and images. For more information, see [Configure Backup Target](PATH).
```

jillian-maroket: @rajeshkio Please use this wording. Replace `PATH` with the actual path to the linked doc section.
cc: @WebberHuang1118

PR #1137, opened 2026-09-18
Remove v1.4 doc from Docusaurus build

jillian-maroket:
## Issue
The Docusaurus build fails with the following error message:

```
[Rspack Deprecation] `experiments.lazyBarrel` config is deprecated and will be removed in Rspack v2.0. Lazy barrel is already stable and enabled by default. Remove this option from your Rspack configuration.
Error: The operation was canceled.
```

I first encountered this issue in PR [#1134](https://github.com/harvester/docs/pull/1134) (created on 2026/9/16). The error message disappeared after I merged `main` into my branch.

## Root cause
@votdev offered the following analysis in [Slack](https://suse.slack.com/archives/C02D9459KFW/p1789564915698889?thread_ts=1789561017.312349&cid=C02D9459KFW):

> The build failure in GitHub Actions is caused by an Out-Of-Memory (OOM Killer) condition (`exit code 137` / SIGKILL), which the runner reports as `The operation was canceled`.
> - GitHub-hosted `ubuntu-latest` runners have a strict memory ceiling of 7 GB RAM.
> - With v1.9 added alongside v1.4, Docusaurus is trying to compile 7 doc versions simultaneously (current/dev + v1.9 through v1.4). Rspack (native memory) together with Node.js V8 heap exceeds the runner's available memory.

## Mitigation
@votdev suggested retiring v1.4 from the active build (the files stay preserved under `versioned_docs/`, but are omitted from the build config). I initially pushed back because of the following reasons:

- We typically include the latest EOL doc in the build.
- The EOL date for v1.5.x (according to the [Support Matrix](https://www.suse.com/suse-harvester/support-matrix/all-supported-versions/harvester-v1-5-x/) is 30 December 2026, so I cannot apply the `EOL` label yet.

However, the error occurred again today, so I have decided to remove v1.4 (with @Vicente-Cheng's consent).

Screenshot 2026-09-18 at 1 10 49 PM


Vicente-Cheng (approved):

votdev:
I switched to draft to prevent merging before tests on https://github.com/harvester/docs/pull/1139 are finished. Using more swap might prevent us from removing the 1.4 version atm.

votdev:
I would suggest to keep that PR open for a while if it comes out that the [increased swap memory](https://github.com/harvester/docs/pull/1139) does not fully fix the issue. If we do not realize problems for some time, we can safely close this PR.

PR #1138, opened 2026-09-18
The note is not required now as the support for Harvester CSI driver RWX is present

saurabhsadhalesuse:
#### Problem:
The note present in the documentation causes confusion for RWX support for harvester-csi driver. 

#### Solution:
This PR removes those notes about RWX support not present.


docs/rancher/csi-driver.md
```diff
@@ -59,12 +59,6 @@ The Harvester CSI driver supports the following features:
 - The Harvester virtual machines that run as guest Kubernetes nodes are in the same namespace.
 - Starting with v0.1.25, if you install the CSI driver without enabling the Harvester cloud provider, ensure that [Volume Snapshots](./csi-driver.md#volume-snapshots) are properly installed; otherwise, the chart installation will fail.
 
-:::note
-
-Currently, the Harvester CSI driver only supports single-node read-write(RWO) volumes. Please follow the [issue #1992](https://github.com/harvester/harvester/issues/1992) for future multi-node `read-only`(ROX) and `read-write`(RWX) support.
```
brandboat, line 64: Instead of deleting the note entirely, we should update it to clarify the supported access modes.
Currently, the harvester csi driver only supports: 
- ReadWriteMany
- ReadWriteOnce
Others modes like ReadWriteOncePod and ReadOnlyMany (ROX) are not supported. Note that https://github.com/harvester/harvester/issues/1992 covers the RWX implementation, but does not include ROX support.

votdev (approved):
LGTM, however, i agree with https://github.com/harvester/docs/pull/1138#discussion_r4080198604

docs/rancher/csi-driver.md
saurabhsadhalesuse: Let me make the changes and submit again.

